PR #16, opened 2017-10-17
Create Deep Review contribution visualization

dhimmel:
This analysis computes contribution stats using git word-diff

Todo:

- [x] Add conda environment
- [x] Explore joyplot alternative
- [x] Consider word count not character count

The current figure is based on https://doi.org/10.7554/eLife.26726.012, but I removed the squareroot transformation.


agitter:
Very cool.  I have no major recommendations, and I see this is still a work in progress.  The longer names are truncated on the figure.

In the text or caption, I'd like to point out that this only one example of the types of contribution analysis we can do with the git log.  These data tell an important story but don't capture issues, discussion, and review.  We can also link to the deep review contribution page.

Do you think the ~70k characters for me and Casey could reflect formatting changes that affected the diff of the entire manuscript?  It may be a coincidence that our counts are so similar.  A rough estimate in Word shows ~225k total characters in the manuscript.

It isn't central to the points of the meta review, but annotating the timeline would make this even more informative.  We could likely see the effects of particular issues that were open to rally contributors.  The plateau in contributions follows the initial preprint.  It's too bad that we fell behind on reviewing and merging pull requests because it would been nice to show how the bioRxiv posting actually attracted many new contributions.

dhimmel:
Here's the latest. It's a ridgeline plot of total words added up to a point in time. In parenthesis are the total words added per contributor.



In https://github.com/greenelab/meta-review/pull/16/commits/4be436a9f26d8286851dca8d59dc7b83b4065c2a I attempted to remove the effect of relocation. The method is poor, but figuring out how to get accurate deltas has been a big timesink... I think we'll have to live with the current method until something better comes along.

agitter:
That looks great.  I agree that it isn't worth your time to do a more fine-grained diff.  The counts that I spot checked seem approximately correct.

I noticed that we have some users who committed with multiple accounts.  The figure shows Paul Agapow/Paul-Michael Agapow and Greg Way/Gregory Way.

What do you think about adding a blank entry before and after me in the plot?  That way none of the ridgelines would overlap.  Could we accomplish that by inserting a null user with no contributions?  I didn't check the plotting code to see if that is feasible or advisable.

dhimmel:
How about now @agitter?

dhimmel:
Don't merge yet. I want to update the environment tomorrow.

agitter (approved):
This looks great to me.  Merge whenever you're ready.

dhimmel:
@agitter I rebased and added the figure to the manuscript.

Waiting on https://github.com/conda-forge/r-gdtools-feedstock/pull/1 so we can export the image as SVG and use that for the manuscript embed. Perhaps review the caption for now (https://github.com/greenelab/meta-review/pull/16/commits/e09a73d037f354340d72a5b0f27917e5342f7050), so this is ready to go when the conda-forge package update occurs.

content/02.main-text.md
```diff
@@ -98,10 +98,17 @@ The entire process can be orchestrated through GitHub with a web browser if a co
...
 In addition, authors are associated with their commits, which makes it easy for contributors to receive credit for their work and helps prevent ghostwriting [@doi:10.1371/journal.pmed.1000023].
-The GitHub contributors page summarizes all edits and commits from each author, providing aggregated information that is not available on other collaborative writing platforms.
+Figure @fig:contrib and the GitHub [contributors page](https://github.com/greenelab/deep-review/graphs/contributors) summarize all edits and commits from each author, providing aggregated information that is not available on other collaborative writing platforms.
 Because our writing process, like others backed by the open git version control system (including Overleaf and Authorea), tracks the complete commit history, it also enables detailed retrospective contribution analysis.
 `TODO: confirm Overleaf and Authorea provide this type of git integration versus something more coarse`
 
+![
+**Deep Review contributions by author over time.**
```
agitter, line 106: I think we've been using generic terms like "open review" and "review manuscript" and haven't referred to this as "Deep Review".  We should introduce that term early on if you want to use it here.

agitter (commented):
This looks good, only one minor comment about the caption.

We could also reference this figure in the second to last paragraph.
> After releasing the first version of our collaborative review in May 2017 [@doi:10.1101/142760], new authors have contributed text (Figure @fig:contrib)

I removed the new author count because I previously included unmerged pull requests.

PR #20, opened 2017-10-21
Add abstract

agitter:
I'd like to keep this short and avoid being too redundant with the existing intro.

### DIFF
--- a/content/01.abstract.md
+++ b/content/01.abstract.md
@@ -1,3 +1,7 @@
 ## Abstract {.page_break_before}
 
-
+Scientific literature reviews are ideal for open, collaborative writing.
+Allowing any interested individual to contribute as an author can strengthen a review, providing broad and fair coverage of the subject matter.
+However, the traditional multi-author writing process breaks down at scale.
+We present techniques for overcoming the challenges of open manuscript writing.
+These include our new software for automating citation and manuscript building as well as approaches for managing distributed authors.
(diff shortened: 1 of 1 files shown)


content/01.abstract.md
```diff
@@ -1,3 +1,7 @@
 ## Abstract {.page_break_before}
 
-
+Scientific literature reviews are ideal for open, collaborative writing.
+Allowing any interested individual to contribute as an author can strengthen a review, providing broad and fair coverage of the subject matter.
+However, the traditional multi-author writing process breaks down at scale.
+We present techniques for overcoming the challenges of open manuscript writing.
```
dhimmel, line 6: Very strong first four sentences!

Do we want to mention the deep review for context? Something like "based on our experience with the Deep Review, where XX authors combined expertise to review deep learning in precision medicine.

content/01.abstract.md
```diff
@@ -1,3 +1,7 @@
...
+These include approaches for managing distributed authors and our new software for automating citation and manuscript building.
```
dhimmel, line 7: I think the abstract needs to mention Manubot by name and provide some more information on it. One point after all is for this article to be the publication citation for the Manubot. I'm happy to propose some text here... probably once we merge this PR (the content that is here is good and mergeable).

dhimmel (approved):

content/01.abstract.md
agitter: My inclination was to keep it as short as possible.  I read several feature article abstracts and some were only 3 sentences.

content/01.abstract.md
agitter: Sounds good.  I'll merge this so you can take a pass over the abstract.

PR #35, opened 2018-02-20
Change timestamp paragraph to aid interpretability

### LINKED ISSUE #32: Bitcoin section may not be ideal for a general audience.
In #26 @agitter noted that the bitcoin/blockchain section may not be interpretable to a general audience. I can take a stab at revising this section after #26 is merged.

cgreene:
A couple things:

* Doh! I cloned the greenelab repo so this is going to create a branch we'll need to prune.
* I tried to provide a bit of context around the need for timestamps.
* I tried to make the section interpretable to someone who is a part of the blockchain lay audience.

This would close #32

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -110,7 +110,7 @@ We implemented support for [OpenTimestamps](https://opentimestamps.org/) in Manu
 This procedure allows one to retrospectively prove that a manuscript version existed prior to its blockchain-verifiable timestamp [@arxiv:1502.04015v1; @url:https://www.bgcarlisle.com/blog/2014/08/25/proof-of-prespecified-endpoints-in-medical-research-with-the-bitcoin-blockchain/; @url:http://blog.dhimmel.com/irreproducible-timestamps/; @url:http://git.dhimmel.com/bitcoin-whitepaper/].
 For Manubot manuscripts, scientific precedence can now be indisputably established, and timestamps protect against attempts to rewrite a manuscript's history.
 Such timestamping practices help ensure accurate histories, potentially alleviating certain authorship or priority disputes.
-Transactions on the Bitcoin blockchain incur a cost that cannot be readily predicted [@url:https://arstechnica.com/tech-policy/2017/12/bitcoin-fees-are-skyrocketing/].
+Since all bitcoin transactions are competing for limited space on the blockchain, the fees required to send a single transaction can be high.
 OpenTimestamps avoids this fee by encoding many timestamps into a single Bitcoin transaction [@url:https://petertodd.org/2016/opentimestamps-announcement], but there can be a lag of a few hours before the transaction is made.
 We judged this to be suitable for the purposes of scientific writing.
 
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -103,15 +103,16 @@ When this repository changes, Travis CI [rebuilds](https://travis-ci.org/greenel
 If successful, the output is deployed back to GitHub (to dedicated [`output`](https://github.com/greenelab/meta-review/tree/output) and [`gh-pages`](https://github.com/greenelab/meta-review/tree/gh-pages) branches).
 As a result, https://greenelab.github.io/meta-review stays up to date with the latest HTML manuscript.
 
-Manubot uses [OpenTimestamps](https://opentimestamps.org/) to timestamp the HTML and PDF outputs on the Bitcoin blockchain before deploying to GitHub.
+The idea of the "priority of discovery" is important to science, and Vale and Hyman [@doi:10.7554/eLife.16931] discuss the importance of both disclosure and validation.
+In their framework, disclosure occurs when a scientific output is released to the world.
+However, for a manuscript that is shared as it is written, being able to establish priority could be challenging.
+We implemented support for [OpenTimestamps](https://opentimestamps.org/) in Manubot to timestamp the HTML and PDF outputs on the Bitcoin blockchain.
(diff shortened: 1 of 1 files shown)


content/02.main-text.md
```diff
@@ -103,15 +103,16 @@ When this repository changes, Travis CI [rebuilds](https://travis-ci.org/greenel
...
-Therefore, timestamping can help ensure accurate manuscript histories, potentially alleviating certain authorship disputes.
-Note that OpenTimestamps is highly scalable and free to use since it encodes many timestamps into a single Bitcoin transaction.
-However, as a result, timestamps are reliant on third-party calendar services until they are upgraded at which point their lineage to the blockchain becomes self-contained.
+For Manubot manuscripts, scientific precedence can now be indisputably established, and timestamps protect against attempts to rewrite a manuscript's history.
+Such timestamping practices help ensure accurate histories, potentially alleviating certain authorship or priority disputes.
+Transactions on the Bitcoin blockchain incur a cost that cannot be readily predicted [@url:https://arstechnica.com/tech-policy/2017/12/bitcoin-fees-are-skyrocketing/].
+OpenTimestamps avoids this fee by encoding many timestamps into a single Bitcoin transaction[@url:https://petertodd.org/2016/opentimestamps-announcement], but there can be a lag of a few hours before the transaction is made.
```
dhimmel: missing space before reference

content/02.main-text.md
```diff
@@ -103,15 +103,16 @@ When this repository changes, Travis CI [rebuilds](https://travis-ci.org/greenel
 If successful, the output is deployed back to GitHub (to dedicated [`output`](https://github.com/greenelab/meta-review/tree/output) and [`gh-pages`](https://github.com/greenelab/meta-review/tree/gh-pages) branches).
 As a result, https://greenelab.github.io/meta-review stays up to date with the latest HTML manuscript.
 
-Manubot uses [OpenTimestamps](https://opentimestamps.org/) to timestamp the HTML and PDF outputs on the Bitcoin blockchain before deploying to GitHub.
+The idea of the "priority of discovery" is important to science, and Vale and Hyman [@doi:10.7554/eLife.16931] discuss the importance of both disclosure and validation.
```
dhimmel: let's move reference to end of sentence.

dhimmel (commented):
Nice.

dhimmel:
I recently read more on the scalability algorithm used by OpenTimestamps. Quite fascinating.

First an aggregator service takes all timestamps in a certain time period (like a second) and creates a merkle tree. The path to the root of this merkle tree is retuned to the user. Hence, timestamps could have latency up to a second to receive the initial pre-upgraded timestamp. Then the 1-second hash is submitted to a calendar, which commits to creating a transaction that descends from this timestamp.

content/02.main-text.md
```diff
@@ -103,15 +103,16 @@ When this repository changes, Travis CI [rebuilds](https://travis-ci.org/greenel
...
-First, scientific precedence can now be indisputably established.
-Second, timestamps can protect against attempts to rewrite a manuscript's history.
-Since timestamps cannot be backdated, alternative histories would have to be created in advance, which is generally infeasible.
-Therefore, timestamping can help ensure accurate manuscript histories, potentially alleviating certain authorship disputes.
-Note that OpenTimestamps is highly scalable and free to use since it encodes many timestamps into a single Bitcoin transaction.
-However, as a result, timestamps are reliant on third-party calendar services until they are upgraded at which point their lineage to the blockchain becomes self-contained.
+For Manubot manuscripts, scientific precedence can now be indisputably established, and timestamps protect against attempts to rewrite a manuscript's history.
+Such timestamping practices help ensure accurate histories, potentially alleviating certain authorship or priority disputes.
+Transactions on the Bitcoin blockchain incur a cost that cannot be readily predicted [@url:https://arstechnica.com/tech-policy/2017/12/bitcoin-fees-are-skyrocketing/].
```
dhimmel: This sentence is not right. I think what you want to say is something like:

> Since all bitcoin transactions are competing for limited space on the blockchain, the fees required to send a single transaction can be high.

No citation needed.

cgreene: I was trying to say that we couldn't say now what your cost would be in the future. Yours is definitely better. I'll use that 👍

content/02.main-text.md
```diff
@@ -103,15 +103,16 @@ When this repository changes, Travis CI [rebuilds](https://travis-ci.org/greenel
 If successful, the output is deployed back to GitHub (to dedicated [`output`](https://github.com/greenelab/meta-review/tree/output) and [`gh-pages`](https://github.com/greenelab/meta-review/tree/gh-pages) branches).
 As a result, https://greenelab.github.io/meta-review stays up to date with the latest HTML manuscript.
 
-Manubot uses [OpenTimestamps](https://opentimestamps.org/) to timestamp the HTML and PDF outputs on the Bitcoin blockchain before deploying to GitHub.
+The idea of the "priority of discovery" is important to science, and Vale and Hyman discuss the importance of both disclosure and validation [@doi:10.7554/eLife.16931].
```
agitter, line 106: Do this line and the `scientific precedence` phrase below give the impression that we advocate for emphasizing the precise timing of scientific discoveries?  I like the Manubot timestamps very much.  I also like the sentiment from [PLOS](http://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.2005203) and others that we shouldn't give all of the credit to the authors who are first to disclose or demonstrate something.

agitter (approved):
Looks great.  I have one comment, but I'm approving because I'm okay merging as-is if you want to keep this in its current form.

dhimmel (approved):

content/02.main-text.md
dhimmel: Yes. I agree with Andy. It should be hard to backdate or falsify the history of a manuscript. However, that does not mean all credit should go to the first manuscript. I haven't looked into the reference enough to know whether this sentence inappropriately fetishizes precedence.

cgreene: I agree with the concern around highlighting the precise timing. However, being able to say roughly when something existed is valuable for many reasons. For example, if we wanted to figure out if something could be predicted computationally before a computer discovered it. Do you have any suggested wording changes? Or should I merge and let you propose a revision?

agitter: I don't have a preferred wording right now.  Let's merge, and I'll think about it.  Adding another example like declaring a prediction in advance could help.

PR #53, opened 2018-06-14
Rephrase PPPR sentences

dhimmel:
Follows up on https://github.com/greenelab/meta-review/pull/52#discussion_r195462826

Tagging @agitter

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -22,8 +22,8 @@ Although we requested that some authors participate for their specific expertise
 In total, the Deep Review attracted {{deep_review_authors}} authors from 20 different institutions (and counting) who were not determined in advance.
 
 Writing review articles in a public forum allows review authors to engage with the original researchers to clarify their methods and results and present them accurately, as exemplified [here](https://github.com/greenelab/deep-review/issues/213).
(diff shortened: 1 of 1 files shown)


dhimmel:
There should be a good PPPR citation that fits here as well

Here is an okay one https://doi.org/10.12688/f1000research.7968.1. Will look for better

agitter (approved):

agitter:
@dhimmel I'm looking for PPPR references and believe I have a good one.  http://doi.org/10.1629/uksg.245 includes a section on challenges and the line
> Due to the large workloads scientists have, there is little motivation for them to participate in PPPR, particularly if their contributions to a paper are not acknowledged by the authors and acted upon.

agitter:
I may also add a more general PPPR reference like:
- https://doi.org/10.1016%2Fj.tig.2015.03.006
- https://doi.org/10.1371%2Fjournal.pmed.1001772
- https://doi.org/10.3389/fncom.2012.00063

dhimmel:
@agitter I liked all your references, check out how I integrated them. Is citing all 4 excessive?

agitter:
That's exactly what I was thinking for the reference placement.  4 may be excessive, but it is easy to cut 1-2 later if we want to.

PR #56, opened 2018-06-14
 Update Deep Review contrib stats to manuscript v1.0

### LINKED ISSUE #51: Re-run Fig 1 up to V1.0 tag (March 6).

dhimmel:
Closes https://github.com/greenelab/meta-review/issues/51 updating Deep Review to https://github.com/greenelab/deep-review/commit/63d2468883ea69ad7ad638c39efab0fcbe026298, which was the v1.0 release version.


dhimmel:
Ready for review.

agitter (changes requested):
In `03.contrib-stats.ipynb` do we need to update the dates in cell 9 used to calculate `post_submission_authors` and the text above?  I don't believe it will change the counts, but we can check.

Brett's name is truncated in `deep-review-contribution.pdf` but the png and svg are fine, so this isn't a problem.

Everything else looks great.

dhimmel:
> In 03.contrib-stats.ipynb do we need to update the dates in cell 9 used to calculate post_submission_authors and the text above? I don't believe it will change the counts, but we can check.

So we should change the date range for `post_submission_authors` to starting on 2017-05-28 (date of first submission) and ending on 2018-03-12 (date of v1.0 release)?

dhimmel:
The way `deep_review_post_submission_authors` is used in the manuscript: 

> After releasing the first version of our collaborative review [@doi:10.1101/142760], {{deep_review_post_submission_authors}} new contributors have updated the manuscript (Figure @fig:contrib) and existing authors continue to discuss new literature, creating a living document.

agitter (approved):
Thanks, this looks perfect and is ready to merge.  Now the new author count matches what we see in the ridge plot.  I agree with your choice of the 2018-03-12 end date, and the phrasing for `deep_review_post_submission_authors` is still accurate.  These are new "contributors" even though not all of them were new authors.

Issue #8, opened 2017-09-29
Initial submission to eLife: timeline & updates


agitter:
We are fairly close to having a full draft ready for submission.  @dhimmel and @cgreene do you have any thoughts on a timeline?  The remaining tasks are the Manubot section, addressing the many inline TODOs, and perhaps a figure to illustrate the issue and pull request system.

dhimmel:
@agitter I'll get to work! Hopefully I can knock most of the Manubot content out next week... I begun yesterday!

agitter:
Great!  I think I nagged @cgreene enough to get it on his TODO list also.

Any thoughts on a figure?  I have no graphic design skill so if I draft it, it would have to be screenshot-based.

cgreene:
I have addressed the major TODOs that I think were assigned to me. :smile: I agree that a quick timeline is good.

agitter:
What do both of you think about a figure?  If we skip that, we could finish a full draft within a week.

dhimmel:
> What do both of you think about a figure? If we skip that, we could finish a full draft within a week.

I think we should have a figure or two. Here are areas of interest:

1. screenshots of the contribution life cycle. Pull request / review comments > merged commit > Travis CI build / updated manuscript.
2. citation exposé showing how the Manubot does citations / references
3. Deep Review contribution plot. Here are possibilities: [Rephetio](http://git.dhimmel.com/rephetio-manuscript/#fig:contribution), [Cognoma](https://github.com/cognoma/cognoma#contributing), or a Gource-movie [like this](https://www.youtube.com/watch?v=dTILX-_JzTs).

agitter:
Those are all excellent ideas.  1 is the most essential because it will help readers understand the main idea and contribution.

For 3, are those figures easy to generate automatically from a commit history or does it take some manual effort?  What would the cognoma graphic look like for us because we only have one repo?  Markdown files by users?  I like the static figures better than the video.

agitter:
I'm following up on this figure idea
> screenshots of the contribution life cycle. Pull request / review comments > merged commit > Travis CI build / updated manuscript.

I looked through some of the deep review pull requests.  greenelab/deep-review#286 and greenelab/deep-review#347 stood out as two where there was discussion, user tagging, issue referencing that will illustrate advantages of GitHub.  We could show panels with:

1. Review comments
![image](https://user-images.githubusercontent.com/3532898/32682429-363ceaea-c63a-11e7-82f4-19b817570ccc.png)
[...]

PR #59, opened 2018-06-15
Add workflow figure

### LINKED ISSUE #40: Make github workflow figure
#8 has some parts about PR commenting. This would be more general. I'll start with a sketch and then we can discuss improvements.

### LINKED ISSUE #41: Identify specific deep review example
We want a contribution from reader identifies change -> file issue & discuss -> file PR & discuss -> final revisions to PR -> final merging.

@agitter do you want to come up with one of these examples that you particularly like as an exemplar workflow?

agitter:
Closes #40, closes #41.

Should the example issue and pull request from #41 be added to the workflow figure caption?


content/02.main-text.md
```diff
@@ -42,12 +42,21 @@ The official version of the manuscript is _forked_ by individual contributors.
...
 
+![
+**Deep Review editing workflow.**
+Any reader can become a contributor by proposing a change through a pull request.
+In this example, the contributor opens an issue to discuss a manuscript modification.
+A maintainer and additional participant provide feedback, and the maintainer recommends creating a pull request to update the text.
+The contributor creates the pull request, it is reviewed by a maintainer and a participant, and the contributor updates the pull request in response.
+Once the pull request is approved, the maintainer merges the changes into the official version of the manuscript.
+](images/workflow.svg){#fig:workflow width="100%"}
+
```
cgreene: I'd put the specific example here. General case described. Then specific example. Then going into the bit below about comparing unrestricted v. highly structured editing.

content/02.main-text.md
```diff
@@ -42,12 +42,21 @@ The official version of the manuscript is _forked_ by individual contributors.
...
-Conversations during review can reference other pull requests, issues, or authors, linking the relevant people and content.
+Conversations during review can reference other pull requests, issues, or authors, linking the relevant people and content, as illustrated in Figure @fig:workflow.
 Reviewing batches of revisions that focus on a single theme is more efficient than independently discussing isolated comments and edits, and it helps maintain consistent content and tone across different authors and reviewers.
 Once all requested modifications are made, the manuscript maintainers, a subset of authors with elevated GitHub permissions, formally approve the pull request and merge the changes into the official version.
-`TODO: need a figure with a flowchart showing this process`
+The Deep Review [issue](https://github.com/greenelab/deep-review/issues/575) and [pull request](https://github.com/greenelab/deep-review/pull/638) on protein-protein interactions demonstrate this process in practice.
```
cgreene: I'd remove this and put things in with more detail below the general case.

cgreene:
Also, do we still want to use https://github.com/greenelab/meta-review/issues/8#issuecomment-353614428 in some way to show the actual interface? It may be too confusing because that is about yet another type of discussion. Perhaps it'd be better in the manubot section describing features?

If we don't want it either place, we should probably close #8.

content/02.main-text.md
dhimmel: Can we add an HMTL comment that adds the Google Draw URL in the manuscript body.

And/or is there a way to save google draw as a file, so anyone could import it back into Google Draw? I don't see a way, so I think URL in a comment is fine:

```md
<!–– Google Drawing View Link https://docs.google.com/drawings/d/17mst0Z1RMXegeGhM6SwQSJ4nhHgMMh5csyV4MwhsgAY/edit?usp=sharing ––>
```

dhimmel:
Reminder to update to 1.0.3:

> I bolded the "Document" and "Discussion" headers and fixed the alignment of the actions on the right hand side. Now labeled 1.0.3

dhimmel:
> Also, do we still want to use https://github.com/greenelab/meta-review/issues/8#issuecomment-353614428 in some way to show the actual interface? It may be too confusing because that is about yet another type of discussion. Perhaps it'd be better in the manubot section describing features?

I like the following figure:

> manubot-draft

But not sure how much it fits here. Possible we could integrate panel A and B. But I think it would be more trouble than it's worth.

content/02.main-text.md
agitter: By put in more detail, do you mean describe the conversation in the issue and pull request?

content/02.main-text.md
agitter: I couldn't figure out how to link directly to the image on Google Drawing.  I added this comment in 9fe7285.

agitter:
Now that we have such a strong workflow figure, I think the older figure in #8 is less essential.  I still like it but would be willing to skip it if we want to focus our efforts elsewhere.

However, the comment https://github.com/greenelab/meta-review/issues/8#issuecomment-334589624 reminded me that we might want to add the build process to the workflow figure.  Should we show Travis generating the update manuscript after the merge?  Or a Travis build failing and prompting the contributor to fix something?

agitter:
e5fa68d11523ed5f9d2c7a350910b840def555f6 adds version 1.0.3 of the figure.

9f749f7dbbef78da658dfef01892380981b04528 adds a longer description of the Deep Review workflow example.  Is this what you had in mind @cgreene?  Should I use real names in addition to the generic "contributor", "maintainer", and "participant"?

cgreene (approved):
With the changes this LGTM 👍

agitter:
*Commented on the wrong pull request - deleted*

agitter:
When I build the manuscript locally, the Figure reference is broken and the caption does not show up:
[manuscript.pdf](https://github.com/greenelab/meta-review/files/2111669/manuscript.pdf)

@dhimmel any ideas?  I may not have the latest `manubot` conda environment.  There also isn't room for the figure and caption on a single page.

content/02.main-text.md
```diff
@@ -42,12 +42,27 @@ The official version of the manuscript is _forked_ by individual contributors.
...
 The process of writing and revising material can be orchestrated through GitHub with a web browser.
 
+![
+**Deep Review editing workflow.**
+Any reader can become a contributor by proposing a change through a pull request.
+In this example, the contributor opens an issue to discuss a manuscript modification.
+A maintainer and additional participant provide feedback, and the maintainer recommends creating a pull request to update the text.
+The contributor creates the pull request, it is reviewed by a maintainer and a participant, and the contributor updates the pull request in response.
+Once the pull request is approved, the maintainer merges the changes into the official version of the manuscript.
+](images/workflow.svg){#fig:workflow width="100%"}
```
dhimmel: Let's reduce the width, since currently the text in the figure is much bigger than other text in the PDF. I wonder if we should do `width="3.5in"` or whatever inch width is most appropriate to make the font size 11 or 12.

dhimmel (commented):
> When I build the manuscript locally, the Figure reference is broken and the caption does not show up

I'm not sure what the issue is. I'd suggest merging after the width is updated and we can deal with any rendering issues if they persist after the CI build's environment.

content/02.main-text.md
agitter: 75% width looked pretty good in one of my local builds so I changed to that in ea62ce6.  Everything else was approved so I'll merge once this build passes.

I suggest we keep #8 open until we finish discussing the GitHub review figure.

dhimmel: I prefer absolute width to relative width. The absolute width should be set to the width that the figure is the right size. I think absolute has some nice advantages when scaling the page IIRC.

PR #60, opened 2018-06-18
Add citation examples table

### LINKED ISSUE #48: concrete citation examples
This would be a nice table. We would have a concrete example of - the DOI/PMID/arXiv style of citation, how it appears in the text, and then how the reference appears.

agitter:
Closes #48.

I did not want to introduce new references for the sake of illustrating PMID and PMCIDs, so I replaced existing DOIs with these identifiers.  As a workaround for escaping `@` , I left it out of the example column.  The table can illustrate the concept without providing the full instructions or syntax.

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -19,7 +19,7 @@ Although we invited some specific experts to participate, most authors discovere
 In total, the Deep Review attracted {{deep_review_authors}} authors who were not determined in advance from 20 different institutions.
 
 Writing review articles in a public forum allows review authors to engage with the original researchers to clarify their methods and results and present them accurately, as exemplified [here](https://github.com/greenelab/deep-review/issues/213).
-Additionally, discussing manuscripts in the open generates valuable post-publication peer review [@pmid:25851694; @doi:10.1371/journal.pmed.1001772; @doi:10.3389/fncom.2012.00063] or pre-publication peer review of preprints [@doi:10.7554/eLife.38532].
+Additionally, discussing manuscripts in the open generates valuable post-publication peer review [@pmid:25851694; @doi:10.1371/journal.pmed.1001772; @doi:10.3389/fncom.2012.00063] or pre-publication peer review of preprints [@tag:Avasthi2018_preprints].
 Because incentives are commonly lacking to provide public peer review of existing literature [@doi:10.1629/uksg.245], open collaborative reviews --- where authorship is open to anyone who makes a valid contribution --- could help spur more post-publication peer review.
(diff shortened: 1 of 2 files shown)


content/02.main-text.md
```diff
@@ -73,14 +73,28 @@ In addition, Manubot relies on extensions from [Pandoc markdown](https://pandoc.
...
+| --- | --- | --- | --- |
+| Digital Object Identifier (DOI) | DOI [Content Negotiation](https://citation.crosscite.org/docs.html) | `doi:10.1098/rsif.2017.0387` | [@doi:10.1098/rsif.2017.0387] |
+| PubMed Identifier (PMID) | NCBI's [E-utilities](https://www.ncbi.nlm.nih.gov/books/NBK25501/) | `pmid:25851694` | [@pmid:25851694] |
+| PubMed Central Identifier (PMCID) | NCBI's [Citation Exporter](https://www.ncbi.nlm.nih.gov/pmc/tools/ctxp/) | `pmcid:PMC4719068` | [@pmcid:PMC4719068] |
+| arXiv identifier | [arXiv API](https://arxiv.org/help/api/index) | `arxiv:1502.04015v1` | [@arxiv:1502.04015v1] |
+| URL | [Greycite](http://greycite.knowledgeblog.org/) [@arxiv:1304.7151v1] | `url:https://lgatto.github.io/open-and-open/` | [@url:https://lgatto.github.io/open-and-open/] |
+| Tag | Source for tagged identifier | `tag:Avasthi2018_preprints` | [@tag:Avasthi2018_preprints] |
+
+Table: Citation types supported by Manubot.
+Authors map each tag to one of the other supported identifier types.
```
cgreene: Should this sway "Authors _may_ map..." I think the citation by permanent ID is a major benefit and making sure that someone scanning the legend doesn't get confused on the optional tag syntax seems helpful.

cgreene (approved):
Just one comment. Other than that, LGTM 👍

content/02.main-text.md
agitter: Updated in e7add1b

agitter:
I should test a local build before merging to confirm the table looks okay.

agitter:
The local build worked so I'm merging.  The table width fits well, even with the long URL.

The PMID example journal name is `Trends in genetics : TIG`, which looked strange but is how it is stored in the NLM catalog.  We can swap it later if anyone objects.

We should move the placement of this example table later.  In the PDF, there is half a page of whitespace now between the paragraph referencing the table and the table, which appears on the next page.

dhimmel:
Looks good to me. Thanks @agitter for making this table!

PR #62, opened 2018-06-18
Absolute workflow figure size and fix caption

### LINKED ISSUE #61: Corrections to workflow figure
Follow up from #59:
- [x] Workflow figure number is not generated
- [x] Workflow figure caption does not appear
- [x] Use absolute instead of relative width (https://github.com/greenelab/meta-review/pull/59#discussion_r196177624)

@dhimmel I missed your comment about the width before merging so I added it here.

agitter:
We needed to have an empty line between the figure and the comment with the Google Drawing link.  Here is my local build with a 5 in absolute width: [manuscript.pdf](https://github.com/greenelab/meta-review/files/2112842/manuscript.pdf)

Closes #61.


dhimmel (approved):
Nice

PR #67, opened 2018-06-26
Add in Reference from #63 for modern data workflow for living data

### LINKED ISSUE #63: Developing a modern data workflow for living data
https://doi.org/10.1101/344804

Should we include this in our discussion?   Continuous integration for data management has parallels to continuous integration for manuscripts.

vsmalladi:
Closes #63
_agitter: edited to close related issue upon merge_

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -1,6 +1,6 @@
 ## Introduction {.page_break_before}
 
-Openness in research --- which includes sharing code, data, and manuscripts --- benefits the researchers who practice it [@doi:10.7554/eLife.16800], their scientific peers, and the public [@doi:10.1101/344804].
+Openness in research --- which includes sharing code, data, and manuscripts --- benefits the researchers who practice it [@doi:10.7554/eLife.16800], their scientific peers [@doi:10.1101/344804], and the public.
 `TODO: more references needed`
 Open scholarly writing, a form of crowdsourcing [@pmcid:PMC4719068], has particular benefits for review articles, which present the state of the art in a scientific field [@doi:10.1371/journal.pcbi.1003149].
 Literature reviews are typically written in private by a closed team of colleagues.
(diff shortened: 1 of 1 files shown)


agitter (commented):
This placement of the reference suggests that the paper describes benefits of the open workflow to the general non-scientist public. Is that the main focus, or is it aimed toward the scientific community?

Before merging, we should also add this reference to the continuous integration paragraph.  I think it fits better there than the "paper of the future" paragraph.

vsmalladi:
Yes, I think I should move the reference to the 'scientific peers' and will add the reference to the continuous integration section.

agitter (approved):
This placement looks okay to me.  We could pull the reference out separately later to describe what this approach does in more detail, but it may be slightly out of scope.

Let's wait for @dhimmel to review before merging.

vsmalladi:
Sounds good.

content/02.main-text.md
```diff
@@ -1,6 +1,6 @@
 ## Introduction {.page_break_before}
 
-Openness in research --- which includes sharing code, data, and manuscripts --- benefits the researchers who practice it [@doi:10.7554/eLife.16800], their scientific peers, and the public.
+Openness in research --- which includes sharing code, data, and manuscripts --- benefits the researchers who practice it [@doi:10.7554/eLife.16800], their scientific peers [@doi:10.1101/344804], and the public.
```
dhimmel: I'm in favor of removing the reference to "Developing a modern data workflow for living data" here. It's a bizarrely specific reference when we don't cite other more general works. We probably should cite more general works, but these points are also rather self evident.

agitter: @dhimmel despite what I wrote above, upon further consideration I agree with you about removing it here.  It fits better in the continuous integration paragraph.  Are you okay adding it in that paragraph?

dhimmel: Yeah I think it is appropriate for the CI sentence... It's a similar use case of CI

agitter:
@vsmalladi will you be able to make this small change soon so we can merge?  @dhimmel's suggestion above will also resolve the merge conflict.

vsmalladi:
@agitter Yes I will be able to edit it now.

content/02.main-text.md
```diff
@@ -119,7 +119,7 @@ Going forward, we hope to integrate Manubot with the larger JATS ecosystem.
 For now however, the primary Manubot output is HTML intended to be viewed in a web browser.
 
 Manubot performs continuous publication: every update to a manuscript's source is automatically reflected in the online outputs.
-The approach uses continuous integration (CI) [@doi:10.1038/550143a; @doi:10.1038/nbt.3780], specifically [Travis CI](https://travis-ci.org/) at the moment, to monitor for changes.
+The approach uses continuous integration (CI) [@doi:10.1038/550143a; @doi:10.1038/nbt.3780; @doi:10.1101/344804, @doi:10.1101/344804], specifically [Travis CI](https://travis-ci.org/) at the moment, to monitor for changes.
```
dhimmel: Has been added twice.

dhimmel (changes requested):

content/02.main-text.md
vsmalladi: @dhimmel Ah thanks for the catch. Will fix now.

dhimmel (approved):

PR #69, opened 2018-06-26
Remove todos and start future paper discussion

agitter:
I'm removing several TODOs and using issues to track open items because they better support discussion.  It also makes the current version of the manuscript more readable.  As part of that, I started a Discussion paragraph that closes #30 and includes some references from #47.

The other issues for these TODOs are #44, #45, #66, and #68.

PR #74, opened 2018-07-13
Manually edit two URL references

agitter:
This adds authors for the two URL references added in #69.

@dhimmel another book reference adding in that pull request also has questionable formatting:
> **Opening Science**_Springer International Publishing_ (2014) https://doi.org/10.1007/978-3-319-00026-8

The editors are extracted correctly, but our CSL file does not show book editors.  Do we want to update the CSL or do you prefer this style?

### DIFF
--- a/content/manual-references.json
+++ b/content/manual-references.json
@@ -165,5 +165,49 @@
         "family": "Himmelstein"
       }
     ]
+  },
+  {
+    "standard_citation": "url:http://blogs.nature.com/naturejobs/2017/06/01/techblog-c-titus-brown-predicting-the-paper-of-the-future",
+    "URL": "http://blogs.nature.com/naturejobs/2017/06/01/techblog-c-titus-brown-predicting-the-paper-of-the-future/",
+    "title": "TechBlog: C. Titus Brown: Predicting the paper of the future",
+    "container-title": "Naturejobs",
+    "issued": {
+      "date-parts": [
+        [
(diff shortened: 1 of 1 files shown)


agitter:
These are references 45 and 48 in this local build: 
[manuscript.pdf](https://github.com/greenelab/meta-review/files/2193781/manuscript.pdf)

dhimmel:
> The editors are extracted correctly, but our CSL file does not show book editors. Do we want to update the CSL or do you prefer this style?

Do you think we want our CSL to display editors if authors do not exist? And if so, always or just in the instance of a book?

dhimmel (approved):

agitter:
> Do you think we want our CSL to display editors if authors do not exist?

I don't have a strong opinion and was wondering whether you do.  The [APA](http://irsc.libguides.com/apa/bookwitheditorrefernceexample) and [MLA](http://irsc.libguides.com/c.php?g=146579&p=963293) styles would include the editors for a book without authors.  I also have a slightly preference to include them.

We could also leave the CLS as-is for now and continue the discussion later in https://github.com/greenelab/manubot-rootstock/issues/53

dhimmel:
> We could also leave the CLS as-is for now and continue the discussion later

Sounds good. I think I'm in favor of falling back to editors, but that will be an upstream modification, so this PR should be good to go.

Issue #65, opened 2018-06-21
meta-review authorship


agitter:
We recently reviewed all manubot and manubot-rootstock issues and pull requests to acknowledge all contributors.  @slochower and @vsmalladi your discussion and implementation (summarized below) substantially contributed to the Manubot design.  We would like to invite you to join as authors of this meta-review manuscript.  This manuscript discusses collaborative writing and Deep Review experience as an example.  It will also serve as the primary Manubot reference.

We are following [ICMJE authorship criteria](http://www.icmje.org/recommendations/browse/roles-and-responsibilities/defining-the-role-of-authors-and-contributors.html#two) so in addition to your previous contributions you would need to
- help draft or revise the manuscript
- approve the final manuscript
- agree to be accountable for the work

Please let us know if you would like to join as authors.  There are several open issues where you could help.

---
Contribution summary as of 2018-06-19
```
slochower:
Versioning discussion https://github.com/greenelab/manubot-rootstock/pull/101
Formatting change https://github.com/greenelab/manubot-rootstock/pull/56
Fix typos https://github.com/greenelab/manubot-rootstock/pull/25
Fix typo https://github.com/greenelab/manubot-rootstock/pull/22
Equation, figure, and table numbering https://github.com/greenelab/manubot-rootstock/pull/8
Setup discussion https://github.com/greenelab/manubot-rootstock/issues/6
Journal compatibility discussion https://github.com/greenelab/manubot-rootstock/issues/4
Web viewing discussion https://github.com/greenelab/manubot-rootstock/issues/58
Math rendering discussion https://github.com/greenelab/manubot-rootstock/issues/55
Build error report https://github.com/greenelab/manubot-rootstock/issues/47
PDF formatting discussion https://github.com/greenelab/manubot-rootstock/issues/20
SVG discussion https://github.com/greenelab/manubot-rootstock/issues/14
Citation dicussion https://github.com/greenelab/manubot-rootstock/issues/2
Automated numbering discussion https://github.com/greenelab/manubot-rootstock/issues/1
Citation errors https://github.com/greenelab/manubot/issues/33
Time zone discussion https://github.com/greenelab/manubot/issues/3

vsmalladi:
Setup instructions https://github.com/greenelab/manubot-rootstock/pull/84
Image width https://github.com/greenelab/manubot-rootstock/pull/79
CSL examples https://github.com/greenelab/manubot-rootstock/pull/46
[...]

PR #75, opened 2018-07-13
Add @slochower and @vsmalladi as authors

slochower:
AFAICT the authorship order is determined by the loop `{% for author in authors %}` which I think will follow the order of the stanzas in `metadata.yaml`. So I inserted myself and @vsmalladi as mentioned in https://github.com/greenelab/meta-review/issues/65#issuecomment-404884954.

I added boilerplate information for @vsmalladi that found on his most recent publication. @vsmalladi please review!


dhimmel:
> I added boilerplate information for @vsmalladi that found on his most recent publication.

Probably we should let @vsmalladi add himself. Our convention is to have authors add themselves to attest to their authorship consent.

slochower:
Makes sense.

slochower:
Makes sense.

agitter:
@slochower can you please delete the file `content/.00.front-matter.md.swp`?

slochower:
Done. (Sorry, flaky SSH connection dropped and presumably left the `vim` swap file that I didn't notice.)

content/metadata.yaml
```diff
@@ -17,6 +17,15 @@ author_info:
...
     funders: GBMF4552
+  -
+    github: slochower
+    name: David R. Slochower
+    initials: DRS
+    orcid: 0000-0003-3928-5050
+    twitter: drslochower
+    email: slochower@gmail.com
+    affilitations:
+      - Skaggs School of Pharmacy and Pharmaceutical Sciences, University of California, San Diego, 9500 Gilman Drive, La Jolla, CA 92093
```
agitter: Can you please keep `Skaggs School of Pharmacy and Pharmaceutical Sciences, University of California, San Diego` and remove the address?

Everything else looks good.

slochower: I was wondering about that... Usually the address is part of the affiliation when I submit my information to journals. I've got no problem removing it now, but should there be an entry for that in the metadata so that it can be included (or not) depending on the template?

slochower: Also along those lines, what about a flag for corresponding author status? The default could be no, but if one author has in their `yaml` block `corresponding: true` then the template could detect it and add an asterisk (or whatever)? Same idea could be used for `contributed equally: true` or something similar? (Obviously these topics could be migrated to an issue or PR on `manubot` if useful.)

slochower: Also along those lines, what about a flag for corresponding author status? The default could be no, but if one author has in their `yaml` block `corresponding: true` then the template could detect it and add an asterisk (or whatever)? Same idea could be used for `contributed equally: true` or something similar? (Obviously these topics could be migrated to an issue or PR on `manubot` if useful.)

agitter: Let's create an issue in manubot-rootstock if we want to modify metadata.yaml.  We should do it upstream so that all derivative manuscripts could take advantage of the updates.

Briefly, an address line does make sense to me.  At one point in deep review https://github.com/greenelab/deep-review/pull/550 I added explicit support for corresponding authors.  We then discussed supporting arbitrary symbols for co-first, corresponding, or other annotations in https://github.com/greenelab/manubot-rootstock/issues/7 and https://github.com/greenelab/manubot-rootstock/pull/10.  That's what we did for [deep review](https://github.com/greenelab/deep-review/blob/5df298135b65c72d164f69571fb303c5d1db21b2/content/metadata.yaml#L20), but it isn't in manubot-rootstock currently.

agitter (approved):

Issue #58, opened 2018-06-14
Ping manubot-rootstock contributors to determine if they want to be acknowledged


cgreene:
See #50 for the approach we used there. Dividing these into two to avoid e-mail noise.

agitter:
Hello manubot-rootstock contributors. We are writing a manuscript describing how manubot was used to write and coordinate our deep learning review. We would like to acknowledge you by name. Please let us know if you approve and, if so, what name you would like us to use.

@vsmalladi @rgieseke @slochower @adebali @petebachant 

@evancofer already responded https://github.com/greenelab/meta-review/issues/50#issuecomment-397383234, but I'm noting his manubot-rootstock contribution here as well.

petebachant:
My contributions were negligible, so no need for acknowledgment. Thanks for asking though--very cool process!

vsmalladi:
I approve. Please use Venkat S. Malladi.

Let me know if I can help with the manuscript at all.

rgieseke:
Sure, thanks! Please use

Robert Gieseke

evancofer:
Thanks! Name to use is Evan M. Cofer

slochower:
Great, thanks. David R. Slochower.

agitter:
Hello manubot-rootstock beta testers. We are writing a manuscript describing how manubot was used to write and coordinate our deep learning review. We would like to acknowledge you by name for your feedback from OpenCon (https://github.com/greenelab/manubot-rootstock/pull/89). Please let us know if you approve and, if so, what name you would like us to use.

@zambujo @RaoOfPhysics @broadwym @schliebs @Hillea

agitter:
Hello manubot-rootstock and manubot contributors. We are writing a manuscript describing how manubot was used to write and coordinate our deep learning review. We would like to acknowledge you by name for your contributions, suggestions, and discussion. Please let us know if you approve and, if so, what name you would like us to use.

@dsiddy @Miserlou @olgabot @jspauld @mlissner @anseljh @Sieboldianus @lierdakil

dsiddy:
Thanks for your consideration in reaching out, @agitter. It's acknowledgement enough for me. 👌

adebali:
Thanks, Ogun Adebali.

lierdakil:
Thanks, but I really didn't contribute much, so I don't think you need to mention me personally. Mentioning pandoc-crossref would be nice if you're still using it though (IIRC you were using it at some point)

[...]

PR #77, opened 2018-07-13
Add acknowledgements for manubot contributors

agitter:
This contains everyone through https://github.com/greenelab/meta-review/issues/58#issuecomment-398861554

### DIFF
--- a/content/99.back-matter.md
+++ b/content/99.back-matter.md
@@ -1,7 +1,9 @@
 ## Acknowledgements
 
 `TODO: deep review authors for support in testing this process`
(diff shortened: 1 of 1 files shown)


agitter:
Updated to include everyone through https://github.com/greenelab/meta-review/issues/58#issuecomment-404923632

I set a low bar for acknowledgement because I think there is no harm in listing someone and it is useful to show the community's engagement with Manubot.

agitter:
There hasn't been any activity in #58 so we could merge this today.

dhimmel (approved):

PR #78, opened 2018-07-13
Fix typo in slochower metadata

agitter:
I missed this when reviewing #75.  I'll merge once the build passes.


slochower:
I actually noticed the same thing. That was my fault.

dhimmel (approved):

Issue #66, opened 2018-06-22
Organize open science discussion and references


agitter:
In the intro and second discussion paragraph, the references about open science were selected arbitrarily.  The current draft has:
> Openness in research — which includes sharing code, data, and manuscripts — benefits the researchers who practice it [1], their scientific peers, and the public.

> Many others have embraced open science principles and piloted open approaches toward drug discovery [38,39], data management [40–42], and manuscript review [43]. 

We should more explicitly consider what we want to include here and search for other relevant work.

agitter:
There are additional ideas and references for open science in https://doi.org/10.7287/peerj.preprints.2711v3

slochower:
> We should more explicitly consider what we want to include here and search for other relevant work.

I just noticed there's a Wikipedia article on Open Science that lists an Open science taxonomy. One thought is to mention the top-level hierarchy -- open access, open data, open reproducible research, open science evaluation, open science policies, and open science tools -- and then add a sentence or two about particularly interesting examples, like drug discovery or data management.

agitter:
There is some good material here.  The primary sources for the six principles --- http://openscienceasap.org/open-science/ and http://www.know-center.tugraz.at/download_extern/papers/open_science.pdf for the first four --- could be useful for introducing them.

The [Arguments_for_open_science](https://en.wikipedia.org/wiki/Open_science#Arguments_for_open_science) section could also provide good references for our introduction.

PR #81, opened 2018-07-18
Add two sentences on openness in research

slochower:
Refs https://github.com/greenelab/meta-review/issues/66

Rework the first sentence and add two more sentences on the specific benefits of open research, with two more citations.

The text is rough, but this is a start.

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -1,6 +1,8 @@
 ## Introduction {.page_break_before}
 
-Openness in research --- which includes sharing code, data, and manuscripts --- benefits the researchers who practice it [@doi:10.7554/eLife.16800], their scientific peers, and the public.
+Openness in research benefits the researchers who practice it [@doi:10.7554/eLife.16800], their scientific peers, and the public.
+Improving the availability of code and data helps ensure the methods are transparent and the results can be reproduced  [@doi:10.1038/s41562-016-0021].
+Publishing in an open access journal increases the number of readers and reviewers [@doi:10.1371/journal.pone.0000308].
 Open scholarly writing, a form of crowdsourcing [@pmcid:PMC4719068], has particular benefits for review articles, which present the state of the art in a scientific field [@doi:10.1371/journal.pcbi.1003149].
 Literature reviews are typically written in private by a closed team of colleagues.
 In contrast, broadly opening the process to anyone engaged in the topic --- such that planning, organizing, writing, and editing occur collaboratively in a public forum where anyone is welcome to participate --- can maximize a review's value.


content/02.main-text.md
```diff
@@ -1,6 +1,8 @@
 ## Introduction {.page_break_before}
 
-Openness in research --- which includes sharing code, data, and manuscripts --- benefits the researchers who practice it [@doi:10.7554/eLife.16800], their scientific peers, and the public.
+Openness in research benefits the researchers who practice it [@doi:10.7554/eLife.16800], their scientific peers, and the public.
```
agitter, line 3: We could add the new article https://www.nature.com/articles/s41467-018-05227-z here: `their scientific peers [@doi:10.1038/s41467-018-05227-z]`

I'm also looking for a reference that discusses why open research benefits the public.  There is a brief statement at https://www.plos.org/open-access/
> Much scientific and medical research is paid for with public funds. Open Access allows taxpayers to see the results of their investment.

but there may be something with more depth.

agitter (commented):
This looks good to me, but I'll wait for @dhimmel before deciding to merge.  See if you like the optional new reference.

content/02.main-text.md
slochower: I saw the point about funding as well. I just wasn't sure how universally it applies to other countries.

content/02.main-text.md
```diff
@@ -1,6 +1,8 @@
...
+Improving the availability of code and data helps ensure the methods are transparent and the results can be reproduced  [@doi:10.1038/s41562-016-0021].
+Publishing in an open access journal increases the number of readers and reviewers [@doi:10.1371/journal.pone.0000308].
```
dhimmel, line 5: This citation (https://doi.org/10.1371/journal.pone.0000308) is about publicly sharing data increasing citations, not about the open access citation advantage.

content/02.main-text.md
dhimmel: > We could add the new article https://www.nature.com/articles/s41467-018-05227-z here

@agitter the article behind that link is [Assessment of the impact of shared brain imaging data on the scientific literature](https://doi.org/10.1038/s41467-018-04976-1). Personally, I find this study severely shortsighted in that it judges the value of a study by the impact of the journal it was published in:

> Several articles that used INDI data have been published in high-impact specialized journals (e.g., Biological Psychiatry and Neuron) and general-interest journals (e.g., Proceedings of the National Academy of Sciences and Nature Communications)

This sort of reasoning is NOT OPEN and runs counter to the longterm vision of Manubot. In the longterm, we want Manubot to enable scholarly publishing & communication without journals. This study just shows that studies that reuse brain imaging data can be published in high impact journals.

dhimmel (commented):
I'd like to flesh out the discussion a bit more and then come back to these two sentences, and see where in the manuscript they fit best, if at all.

My worry is that they currently weaken the introduction, which succinctly argues the benefit of open manuscript preparation for review articles. If we want to discuss open science in more depth at the beginning, we should make sure it leads the reader to the observation that open manuscript drafting is the next frontier of open science. And that open drafting will complement the other progress being made towards open science (open access, open data, open source, article-level metrics, etcetera).

slochower:
> My worry is that they currently weaken the introduction, which succinctly argues the benefit of open manuscript preparation for review articles. If we want to discuss open science in more depth at the beginning, we should make sure it leads the reader to the observation that open manuscript drafting is the next frontier of open science.

Do you think the introduction would benefit from additional sentences on open science / open research, as mentioned in #66?

content/02.main-text.md
agitter: I also did not like the focus on high-impact journals in the original article.  I thought the editorial was more broad and slightly better, but let's skip it.

content/02.main-text.md
slochower: There are better ones. This seems to be one of the most cited articles on the open access citation advantage -- http://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.0040157 -- but the articles it considered are quite old, from 2004-2005. I only skimmed this one, but it is more recent and seems to make the same points (see section 2.1 and Figure 1): https://cdn.elifesciences.org/articles/16800/elife-16800-v1.pdf

agitter:
> Do you think the introduction would benefit from additional sentences on open science / open research, as mentioned in #66?

@dhimmel has some good ideas for this, so we'd like to merge what you have here and let him add to it.

PR #86, opened 2018-07-20
Add deep review repository status

### LINKED ISSUE #42: Incorporate additional Deep Review and/or Manubot stats
Table to show scale of content of the project.

agitter:
Closes #42 

I couldn't figure out how to format the `creation_time_utc` stat to automatically populate the date, so it is currently hard coded.  It looks like Jinja2 requires a filter for this:
- https://support.sendwithus.com/jinja/jinja_time/
- http://jinja.pocoo.org/docs/2.10/api/#custom-filters

@dhimmel are there any other [stats](https://github.com/greenelab/meta-review/blob/master/analyses/deep-review-contrib/deep-review-stats.json) we want to add?  These seemed the most relevant to me to show the scale of the project.


dhimmel:
> are there any other stats we want to add? 

The star count is quite incredible IMO.

> It looks like Jinja2 requires a filter

Yuck. At some point in the future, we could just store the pretty formatted date in the JSON.

agitter:
> Yuck. At some point in the future, we could just store the pretty formatted date in the JSON.

Yes, that would be even easier.

I added the stars in 90c8e0a.

dhimmel (approved):

PR #88, opened 2018-07-20
Table summarizing writing alternatives

### LINKED ISSUE #44: Evaluate features of Manubot alternatives
We should check how the features that we discuss relate to the features existing in Overleaf/Authorea.

Authorea + Bibtex (@cgreene will explore) -> Casey
Overleaf + Bibtex (@agitter will explore) -> 
Google Docs + Paperpile (@cgreene) -> 
Word (Installed) + Ref Management Software (@cgreene)-> 
Word (Office 365) + Ref Management Software (maybe @agitter) -> 
Markdown on Github (@cgreene will fill in)

agitter:
Closes #44 

The open and paid pricing rows are the most awkward and biggest.  Any ideas for simplifying those?

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -91,7 +91,7 @@ After four rounds of reviews and pull request edits, a maintainer merged the cha
 We found that this workflow was an effective compromise between fully unrestricted editing and a more heavily-structured approach that limited the authors or the sections they could edit.
 In addition, authors are associated with their commits, which makes it easy for contributors to receive credit for their work and helps prevent ghostwriting [@doi:10.1371/journal.pmed.1000023].
 Figure @fig:contrib and the GitHub [contributors page](https://github.com/greenelab/deep-review/graphs/contributors) summarize all edits and commits from each author, providing aggregated information that is not available on other collaborative writing platforms.
-Because our writing process tracks the complete commit history through git commits, it enables detailed retrospective contribution analysis.
+Because our writing process tracks the complete history through git commits, it enables detailed retrospective contribution analysis.
 
 ![
 **Deep Review contributions by author over time.**
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -33,12 +33,12 @@ These platforms ideally offer version control, multiple permission levels, or ot
 Although they work well for editing, they lack sufficient features for managing a collaborative manuscript and attributing precise credit, which are important for open writing (Table @tbl:platforms).
 
 | Feature | Manubot | Authorea + BibTeX | Overleaf + BibTeX | Google Docs + Paperpile | Word Online<sup>1</sup> | Markdown on GitHub |
-|--------------------------------------------|---------------------------|----------------------------------------------------|----------------------------------------------------------------------------------------------------|---------------------------------------------------|------------------------------------------------------------------------------------|--------------------|
+|---|---|---|---|---|---|---|
 | Multi-author editing | Yes | Yes | Yes | Yes | Yes | Yes |
 | Propose changes | Yes | No | No | Yes | No | Yes |
 | Continuous integration testing | Yes | No | No | No | No | No |
 | Multi-participant conversation for changes | Yes | No<sup>2</sup>  | No<sup>2</sup> | No<sup>2</sup> | No<sup>2</sup> | Yes |
-| Character-level provenance for text | Yes | No (versions tracked by day) | No (simultaneous edits online can be attributed to a single author) | Requires manual inspection of history | Not after changes are accepted | Yes |
+| Character-level provenance for text | Yes | No (versions tracked by day) | No<sup>3</sup> | Requires manual inspection of history | Not after changes are accepted | Yes |
 | Bibliography management | Yes | Yes | Yes | Yes | No, requires the Word desktop application | No |
 | Cite by common identifiers | Yes | No | No | No | No | No |
 | Editing software | Any text editor | Web interface | Web interface | Web interface | Web interface | Any text editor |
@@ -52,6 +52,7 @@ A summary of features that differentiate Manubot from existing collaborative wri
 Features were assessed on June 15, 2018.
 <sup>1</sup>Additional functionality, such as bibliography management, is available by editing the Word document stored in OneDrive with the paid Word desktop application.
 <sup>2</sup>Conversations about modifications take place on the document as comments, annotations, or unsaved chats.  There is no separate integrated forum for discussing and editing revisions.
+<sup>3</sup>Simultaneous edits online can be attributed to a single author in a git commit.
 {#tbl:platforms}
 
 We adopted standard software development strategies in order to enable any contributor to edit any part of the manuscript but enforce discussion and review of all proposed changes.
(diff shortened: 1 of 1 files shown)


agitter:
The table is quite large in my local build: [manuscript.pdf](https://github.com/greenelab/meta-review/files/2215482/manuscript.pdf)

@slochower please let me know if I missed anything you suggested in #44 or if you have other ideas.

dhimmel:
Do we want to add a "Technical expertise" row or something similar. Basically anything that is not WYSIWYG is technical. Or a WYSIWYG... let me think about this some more (just noting down thought before I have to run.)

slochower:
Sorry to suggest this now, but just looking at your table, it occurs to me that it might look better if transposed. What do you think?

slochower:
> The open and paid pricing rows are the most awkward and biggest. Any ideas for simplifying those?

What about just noting those as Available or Not Available? The actual details probably don't matter for this manuscript anyway and will surely be changed by those services as time goes on.

agitter:
> Sorry to suggest this now, but just looking at your table, it occurs to me that it might look better if transposed. What do you think?

We originally had it transposed in the Google Sheet, but there were so many columns I didn't see how to make it fit.  This layout is 13 rows by 7 columns.

agitter:
> What about just noting those as Available or Not Available? 

What would Available mean?  That a free version is available?

I agree that this would be a good way to simplify the table.  Under the current pricing models you get write a simple manuscript for free on any of these platforms.

slochower:
> We originally had it transposed in the Google Sheet, but there were so many columns I didn't see how to make it fit. This layout is 13 rows by 7 columns.

I see your point. If we were only making a PDF copy, I'd suggest putting it at the end, on its own page, in landscape. But I don't think that would play well with the other formats. I don't have any good ideas at the moment.

agitter:
> Do we want to add a "Technical expertise" row or something similar.

That is a good idea.  It relates to the ease of use noted in #44.  I assume that Google Docs and Word are in a "low expertise" category and the other platforms all require more expertise.

slochower:
> I agree that this would be a good way to simplify the table. Under the current pricing models you get write a simple manuscript for free on any of these platforms.

I've changed my mind and think I'd prefer to omit the pricing rows. I think they add unnecessary complexity to the table and the point is hard to make because of the nuances of the pricing. Like Word Online does most of the things and is free, but the desktop app has better track changes, etc., and costs money, but can be available via subscription or standalone or through work, etc. I also have started to think "Open" and "Private pricing" aren't quite right as far as I can tell. It's more like: there's separate pricing if you're a power user and need lots of storage or other bells and whistles, right?

agitter:
4df64f0 removes the pricing rows and adds a technical expertise row.  This is still subject to change, but I liked these proposed revisions.  I'll make one more quick commit noting that features can vary some platforms based on whether a paid version is used.

slochower:
Great. How do/did you link to the local PR build of the PDF?

agitter:
My usual workflow for checking a pull request like this is to pull the branch on a Linux server (my local machine runs Windows).  Then I run `sh build/build.sh` from the root directory `meta-review` and copy `output\manuscript.pdf` into the GitHub comment (drag and drop).  GitHub hosts the file and links to it, which is likely the step you were curious about.

cgreene:
This LGTM 👍

content/02.main-text.md
```diff
@@ -68,7 +91,7 @@ After four rounds of reviews and pull request edits, a maintainer merged the cha
 We found that this workflow was an effective compromise between fully unrestricted editing and a more heavily-structured approach that limited the authors or the sections they could edit.
 In addition, authors are associated with their commits, which makes it easy for contributors to receive credit for their work and helps prevent ghostwriting [@doi:10.1371/journal.pmed.1000023].
 Figure @fig:contrib and the GitHub [contributors page](https://github.com/greenelab/deep-review/graphs/contributors) summarize all edits and commits from each author, providing aggregated information that is not available on other collaborative writing platforms.
-Because our writing process, like others backed by the open git version control system (including Overleaf and Authorea), tracks the complete commit history, it also enables detailed retrospective contribution analysis.
+Because our writing process tracks the complete commit history through git commits, it enables detailed retrospective contribution analysis.
```
dhimmel: Do we want to say "commit" twice in this sentence?

dhimmel (approved):
Looks good. I'd say merge and if we have additional changes can always make in another PR.

content/02.main-text.md
agitter: Fixed in c0e2a22

dhimmel:
Table looks wacky in my Chromium Linux browser:

table-superscript-issue

However [`manuscript.pdf`](https://greenelab.github.io/meta-review/v/5bb972bdb2fd89757d04b1bafb46661d64858162/manuscript.pdf) is OKAY, so not a pressing issue.

One alternative is unicode superscript numbers if this is a widespread issue.

agitter:
The superscripts aren't quite as bad on Chrome in Windows:
image

The table is still too wide to fit nicely though.

slochower:
> The superscripts aren't quite as bad on Chrome in Windows:

FWIW, I made a local adjustment to the CSS that makes superscripts look substantially better locally. I have not tested on other browsers. Just dropping it here, in case it's useful in the future.

```diff
diff --git a/webpage/github-pandoc.css b/webpage/github-pandoc.css
index e0e5e81..a478197 100644
--- a/webpage/github-pandoc.css
+++ b/webpage/github-pandoc.css
@@ -262,7 +262,7 @@ sup {
 }

 sup {
-    top: -0.5em;
+    top: -0.05em;
 }
```

dhimmel:
> FWIW, I made a local adjustment to the CSS that makes superscripts look substantially better locally.

Did you try completely removing the `sub` CSS? It looks like the [default](https://www.w3schools.com/cssref/css_default_values.asp) is:

```css
vertical-align: sub;
font-size: smaller;
```

From https://www.w3schools.com/cssref/pr_pos_top.asp:

> The top property affects the vertical position of a positioned element

slochower:
No, I didn't try that. I'm not sure what effect that would have. 

But while looking into this, I just saw a strange effect in my rendered HTML. In Chrome 68 on linux, there is a scenario whereby a large PDF will be rendered inside its own box with its own scroll bar. I'm not sure which combination of factors lead to this and I don't think it is Manubot's problem. But interesting and somewhat undesirable.

screenshot from 2018-09-19 10-31-32

dhimmel:
> there is a scenario whereby a large PDF will be rendered inside its own box with its own scroll bar

What markdown / html code did you use to embed the PDF? My experience is PDFs cannot be used like images in web browsers. If you want, that vector figure to act like an image, it needs to be converted to SVG.

slochower:
Yep, just the usual way `Caption`. I think you were right -- better to use SVG in that case.

vincerubinetti:
There are several ways to "fix" this table overflow issue, each with their own caveats. In my cursory assessment, my judgement is that simply having a horizontal scroll bar is the most universal and intuitive solution. Do either of you have a preference on how this should be handled? Particularly, consider mobile viewing:

https://css-tricks.com/accessible-simple-responsive-tables/
https://css-tricks.com/wp-content/uploads/2016/03/types-of-responsive-tables.gif

agitter:
Of those four options, I agree that the scroll bar seems the most flexible.  For a table like this we could not squish it enough.

Is it advisable (or technically possible) to freeze a column?  For this table, for instance, it would be helpful to lock the first column and only scroll the others.  However, I doubt that could supported in the markdown to HTML pandoc conversion.

vincerubinetti:
Yes, column/row freezing should be possible, but I did not consider that to be a viable option because of what you said regarding the markdown. Which column/row should be frozen, if any, would be dependent on the type of data in the column and would thus need to be specifiable in the markdown. The same thing would be true for dynamic row/column wrapping (ie typical "responsive design").

dhimmel:
A scroll bar in the web view would be fantastic. However we have to make sure that the PDF export of the webview via weasyprint still shows the full table.

PR #94, opened 2018-08-02
Minor proofreading corrections

slochower:


content/02.main-text.md
```diff
@@ -20,10 +20,10 @@ GitHub is a platform designed for collaborative software development that is ada
 From the start, we made the GitHub repository public, applying a [Creative Commons Attribution License](https://github.com/greenelab/deep-review/blob/master/LICENSE.md) to the manuscript.
 Next, we encouraged anyone interested to contribute by proposing changes or additions.
 Although we invited some specific experts to participate, most authors discovered the manuscript organically through conferences or social media, deciding to contribute without solicitation.
-In total, the Deep Review attracted {{deep_review_authors}} authors who were not determined in advance from 20 different institutions.
+In total, the Deep Review attracted {{deep_review_authors}} authors who were not determined in advance, from 20 different institutions.
```
dhimmel: It seems to me that there should be either 2 or 0 commas:

> the Deep Review attracted {{deep_review_authors}} authors who were not determined in advance from 20 different institutions.

> the Deep Review attracted {{deep_review_authors}} authors, who were not determined in advance, from 20 different institutions.

Not totally sure though what conventions we've been following. @agitter?

dhimmel (approved):

content/02.main-text.md
slochower: I could see the case for two commas. Zero commas confused me because I mentally grouped "in advance from 20" and had to do a double reading.

agitter: @dhimmel's two commas version reads most naturally to me.  I agree that the zero comma version is ambiguous.

agitter (approved):

PR #103, opened 2018-10-10
Add back-matter paragraph to replace front-matter funding info

dhimmel:
At some point, we should probably create an in-body paragraph regarding funding, as I'm guessing it'll be eventually required by a journal. At that point, we can remove the frontmatter funding snippets (or keep them if we don't mind the redundancy).

### DIFF
--- a/content/99.back-matter.md
+++ b/content/99.back-matter.md
@@ -7,7 +7,8 @@ We are grateful for additional Manubot discussion and testing by Alexander Dunke
 
 ## Funding
 
-This work was supported by the Alfred P. Sloan Foundation in [Grant G-2018-11163](https://sloan.org/grant-detail/8501) to DSH and the Gordon & Betty Moore Foundation in [Grant GBMF4552](https://www.moore.org/grant-detail?grantId=GBMF4552) to CSG.
+DSH and CSG were supported by [Grant G-2018-11163](https://sloan.org/grant-detail/8501) from the Alfred P. Sloan Foundation and [Grant GBMF4552](https://www.moore.org/grant-detail?grantId=GBMF4552) from the Gordon & Betty Moore Foundation.
+VSM was supported by [Grant RP150596](http://www.cprit.state.tx.us/files/funded-grants/RP150596.pdf) from the Cancer Prevention and Research Institute of Texas.
 
 ## References {.page_break_before}
 
--- a/content/99.back-matter.md
+++ b/content/99.back-matter.md
@@ -7,7 +7,7 @@ We are grateful for additional Manubot discussion and testing by Alexander Dunke
 
 ## Funding
 
-DSH and CSG were supported by [Grant G-2018-11163](https://sloan.org/grant-detail/8501) from the Alfred P. Sloan Foundation and [Grant GBMF4552](https://www.moore.org/grant-detail?grantId=GBMF4552) from the Gordon & Betty Moore Foundation.
+DSH and CSG were supported by [Grant G-2018-11163](https://sloan.org/grant-detail/8501) from the Alfred P. Sloan Foundation and [Grant GBMF4552](https://www.moore.org/grant-detail?grantId=GBMF4552) from the Gordon and Betty Moore Foundation.
 VSM was supported by [Grant RP150596](http://www.cprit.state.tx.us/files/funded-grants/RP150596.pdf) from the Cancer Prevention and Research Institute of Texas.
 
 ## References {.page_break_before}
--- a/content/99.back-matter.md
+++ b/content/99.back-matter.md
@@ -1,9 +1,13 @@
-## Acknowledgements
+## Acknowledgments
 
 We would like to thank the authors of the Deep Review who helped us test collaborative writing with Manubot.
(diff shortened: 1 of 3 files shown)


content/metadata.yaml
```diff
@@ -16,7 +16,7 @@ author_info:
     email: daniel.himmelstein@gmail.com
     affiliations:
       - Department of Systems Pharmacology and Translational Therapeutics, University of Pennsylvania
-    funders: GBMF4552
+    funders: G-2018-11163 and GBMF4552
```
agitter, line 19: Do you want to preface the new grant number with `Alfred P. Sloan Foundation`?  Some of the other grant numbers have the funder abbreviation, but this one is hard to identify on its own.

agitter:
Once we add a funding paragraph, we can remove the funding information from the author list.  However, this could wait until the later stages of review if we prefer showing funding alongside the authors.

dhimmel:
> Once we add a funding paragraph, we can remove the funding information from the author list.

Let's just add the funding acknowledgment now (see https://github.com/greenelab/meta-review/pull/103/commits/fbf972315c5978fa9551b8c3ee0dbddcdfbd1c1b). We could either delete the metadata.yaml funders fields or remove the jinja2 funders insertion template.

@vsmalladi is there a way you'd like to acknoweldge CPRIT RP150596?

content/metadata.yaml
dhimmel: Yeah I am become less & less of a fan of the way metadata.yaml handles funding. It seems like it should be a list like affiliations. Also funding acks are particular enough that there may not be much automated stuff we can do.

Anyways, let's add a funding paragraph to the manuscript and take it from there.

vsmalladi:
@dhimmel We usually add it to Acknowledgements section as:

V.S.M was supported by Cancer Prevention and Research Institute of Texas (RP150596).

dhimmel:
> V.S.M was supported by Cancer Prevention and Research Institute of Texas (RP150596).

Got it. @agitter there are two separate aspects: which people were grant recipients of a grant used to fund manubot VERSUS which people received support from a grant (potentially not directly to them) that was put towards manubot. How do you usually deal with this distinction?

agitter:
> We could either delete the metadata.yaml funders fields or remove the jinja2 funders insertion template.

Let's remove the template insertion and keep the metadata.yml funders.  I like having the authors' self-declared funding there so we can confirm the new funding statement is accurate.

> How do you usually deal with this distinction?

That's not exactly how I typically categorize funding support.  I would organize funding support based on 1) which funding sources supported the authors' effort on the research being presented and 2) which funding sources supported other resources for the research.  For each authors' effort, it doesn't matter to me whether the author was the grant recipient.  It is common for training grants or large center grants that the grant PI is not directly involved in all research and would not be an author on all manuscripts.  They receive "credit" for the research when reporting all research activities supported by that grant.  As long as the grant number is reported in the manuscript and the author funded by the grant notifies the PI that grant supported that manuscript, it is sufficient.

For other resources, we may have additional funding section statements that funding source ABC supported aspect XYZ of the study.  Those could be computing resources, software development efforts of contributors who are not authors, etc.

It also sounds like we want a new manubot-rootstock issue to discuss funding more generally.  For deep review I had started to automate parts of the funding statement, but it was not sufficiently general.  We can decide where funding information should go and whether any of it should be automated or if it is so custom that it need to be free text in the back matter.

content/99.back-matter.md
```diff
@@ -1,10 +1,14 @@
...
+## Acknowledgments
 
 We would like to thank the authors of the Deep Review who helped us test collaborative writing with Manubot.
 The authors who responded favorably to being acknowledged are Paul-Michael Agapow, Amr M. Alexandari, Brett K. Beaulieu-Jones, Anne E. Carpenter, Travers Ching, Evan M. Cofer, Dave DeCaprio, Brian T. Do, Enrico Ferrero, David J. Harris, Michael M. Hoffman, Alexandr A. Kalinin, Anshul Kundaje, Jack Lanchantin, Christopher A. Lavender, Benjamin J. Lengerich, Zhiyong Lu, Yifan Peng, Yanjun Qi, Gail L. Rosen, Avanti Shrikumar, Srinivas C. Turaga, Gregory P. Way, Laura K. Wiley, Stephen Woloszynek, Wei Xie, Jinbo Xu, and Michael Zietz.
 In addition, we thank Ogun Adebali, Evan M. Cofer, and Robert Gieseke for contributing to the Manubot template manuscript.
 We are grateful for additional Manubot discussion and testing by Alexander Dunkel, Ansel Halliburton, Achintya Rao, and other GitHub users.
 
+## Funding
+
+This work was supported by the Alfred P. Sloan Foundation in [Grant G-2018-11163](https://sloan.org/grant-detail/8501) to DSH and the Gordon & Betty Moore Foundation in [Grant GBMF4552](https://www.moore.org/grant-detail?grantId=GBMF4552) to CSG.
```
agitter: Per https://github.com/greenelab/meta-review/pull/103#issuecomment-431373554, I find it more natural to match which authors were supported by which funding sources and ignore the grant PIs.

dhimmel:
@agitter and @vsmalladi: does this look good to go? @agitter feel free to merge (or touch up then merge, if so).

vsmalladi (approved):
Looks good to me.

agitter (approved):

Issue #97, opened 2018-08-24
Revisions before next submission


agitter:
- [x] Edit the abstract to de-emphasize literature reviews and emphasize open collaboration and automation
- [x] Discuss the new Manubot features that will be included in the v0.2.0 release
- [x] Prune some of the retrospective deep review discussion to focus more on how others can use Manubot
- [x] Add Kipoi as a Manubot use case

agitter:
I added ` Add Kipoi as a Manubot use case` so that we can briefly mention https://github.com/kipoi/website/issues/78  The Kipoi model list [here](https://kipoi.org/groups/) now uses Manubot to extract author information from the DOI or arXiv links.

This could fit with our new discussion of the v0.2.0 features or at the very end of the Manubot section that discusses current usage.

agitter:
Now that #99 is merged, we could close this.  Or I'll make a proofreading pull request and close this issue with that new pull request.

vsmalladi:
@agitter I think making it a proofreading pull request makes sense.

PR #105, opened 2018-10-11
Add Kipoi use case

agitter:
Addresses another item from #97.

I wasn't sure how to phrase this:
- `Manubot's citation utility` (current option)
- `Manubot's citation API`
- `Manubot's citation features`


dhimmel:
I would go with:

1. Manubot's citation API
2. Manubot's citation functionality

I slightly prefer 1, but if API is too technical than I prefer functionality.

dhimmel (approved):

agitter:
Yes, I was worried about API being to technical.  I'll switch to functionality.

PR #112, opened 2018-12-18
Upgrade Manubot Rootstock on 2018-12-18

dhimmel:
Includes enhanced citation support.


agitter (approved):
I'm approving the changes, but do we need to wait for datacite to come back online before merging?  Or we could merge this and immediately switch to a manual or URL-based citation for that reference in a new pull request.

dhimmel:
`@url:https://zietzm.github.io/Vagelos2017/` will not work because there is both a translation-server issue at https://github.com/zotero/translation-server/issues/65 and Greycite is down.

I'll just set the doi metadata in manual references in this PR, using one of the past output references.json as a template.

agitter:
> I'll just set the doi metadata in manual references in this PR, using one of the past output references.json as a template.

Looks good to me.

PR #153, opened 2019-02-09
Update abstract

### LINKED ISSUE #115: "correctness and fairness" in abstract are unclear.
from reading the manuscript, it is clear that fairness means fairness in contribution tracking, but this is not clear in the abbreviated language used in the abstract.

I am somewhat unclear on what "ensure correctness" means in the abstract.

agitter:
Closes #115.  I need to update the response letter before merging but want to agree on the language we should use here first.

@dhimmel "correctness" in the abstract referred to the continuous integration testing.  For example, authors won't have out of sync reference libraries, use invalid formatting, or cite non-existing references.  Is this a good way to clarify that?

### DIFF
--- a/content/01.abstract.md
+++ b/content/01.abstract.md
@@ -2,6 +2,6 @@
 
 Open, collaborative research is a powerful paradigm that can immensely strengthen the scientific process by integrating broad and diverse expertise.
 However, traditional research and multi-author writing processes break down at scale.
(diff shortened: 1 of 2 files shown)


dhimmel (approved):

agitter:
Because the new abstract was approved, I'll update the response letter and merge.

Issue #146, opened 2019-02-08
Reviewer 2 comment 17


agitter:
> `they lack sufficient features for managing a collaborative editing` → This is a bit of an overstatemeent because these tools are used daily by a lot of people to do just that. I agree on the second part on the 'precise credit' aspect but as the sentence is currently written, you seem to imply that even collaborative editing is not possible.

PR #156, opened 2019-02-10
Wording about other collaborative editing platforms

slochower:
I'm not sure whether this fully addresses the concern in #146.


agitter:
Emphasizing the branches may help readers who do software development and understand what that implies, but we may want to be even more explicit to address #146.  We should keep the phrase about attribution of credit.  The other aspect to highlight would be that Manubot/GitHub closely integrate project management (issues, pull requests, etc.) with content and help orchestrate editing and reviewing large lists of thematically related changes.

agitter (changes requested):
This additional sentence should clarify what we mean.  The text looks good to me.

Can you please update the response to reviewers as well before we merge?

slochower:
Ready for re-review.

agitter (approved):
Looks good.  In other responses, I often quote the revised text from the updated manuscript.  However, because we're directly linking to the pull request that may be redundant.

dhimmel (approved):

Issue #149, opened 2019-02-08
Reviewer 2 comment 20


agitter:
> Figure .2 → This could be probably adapted as a raster plot where a dot is painted for each contribution (at the time of contribution). Line would be the same height and the legend would summarize the contribution.

slochower:
In retrospect, I agree and think this figure may be too large. In fact, I think I would favor replacing the author names altogether with "Author ###" or whatever. It's more important to me to see that 30 (?) different authors contributed to Deep Review throughout the 2+ years of development. Since I'm somewhat familiar with a few of the names on the list, I find their actual names to be a bit distracting.

dhimmel:
Hmm, I think the names are a nice touch and add ways to use and interpret the figure. Perhaps they can be made smaller by switching to initials or last names only.

If I have time, I'll see if I can create a plot the same axes as the current plot, but with dots whose area signifies the contribution. It would lose the cumulative aspect, such that at any point in time you could see the contribution distribution of the manuscript at that point. I can't find any examples of these plots online, but I think they have a name (that's not raster plot, maybe [dot plot](http://www.pygal.org/en/latest/documentation/types/dot.html)?)

slochower:
> Hmm, I think the names are a nice touch and add ways to use and interpret the figure. Perhaps they can be made smaller by switching to initials or last names only.

Sure, but consider this low priority since it's totally subjective and only aesthetic :)

> I can't find any examples of these plots online, but I think they have a name (that's not raster plot, maybe [dot plot](http://www.pygal.org/en/latest/documentation/types/dot.html)?)

I imagine so, yeah.

agitter:
I also liked the cumulative aspect of the current visualization.  I don't see the benefits of the dot plot.

agitter:
Recent discussions of Stencila, binder, and related tools reminded me of this reviewer request.  @dhimmel already used a Jupyter notebook to generate the figure.  I believe it would not be too difficult to offer a runnable version of this figure's notebook through binder.  That would help position Manubot in the "manuscript of the future" discussions.

If we like the cumulative aspects of this figure, we could also keep the current version in the manuscript (maybe switching to first initials for author names) but offer an alternative version in the notebook.  Or at least we can point out that the notebook can easily support alternative visualizations of this data.

[...]

PR #161, opened 2019-03-02
Add binder support for deep review contributions figure

agitter:
This follows up on our binder discussion in #149.  To avoid duplicating the conda environment file for the notebooks, I used a symlink from the `binder` directory.  That and the `postBuild` file appear to work.

However, the notebook is not yet executable in binder.  You can test it and see the errors in my fork: https://mybinder.org/v2/gh/agitter/meta-review/binder?filepath=analyses/deep-review-contrib/02.contrib-viz.ipynb

Jupyter throws a kernel error
kernel_error

<details><summary>Stack trace</summary>

```python
Traceback (most recent call last):
  File "/srv/conda/lib/python3.6/site-packages/tornado/web.py", line 1592, in _execute
    result = yield result
  File "/srv/conda/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/srv/conda/lib/python3.6/site-packages/tornado/gen.py", line 1141, in run
    yielded = self.gen.throw(*exc_info)
  File "/srv/conda/lib/python3.6/site-packages/notebook/services/sessions/handlers.py", line 73, in post
    type=mtype))
  File "/srv/conda/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/srv/conda/lib/python3.6/site-packages/tornado/gen.py", line 1141, in run
    yielded = self.gen.throw(*exc_info)
  File "/srv/conda/lib/python3.6/site-packages/notebook/services/sessions/sessionmanager.py", line 79, in create_session
    kernel_id = yield self.start_kernel_for_session(session_id, path, name, type, kernel_name)
  File "/srv/conda/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/srv/conda/lib/python3.6/site-packages/tornado/gen.py", line 1141, in run
    yielded = self.gen.throw(*exc_info)
  File "/srv/conda/lib/python3.6/site-packages/notebook/services/sessions/sessionmanager.py", line 92, in start_kernel_for_session
    self.kernel_manager.start_kernel(path=kernel_path, kernel_name=kernel_name)
  File "/srv/conda/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/srv/conda/lib/python3.6/site-packages/tornado/gen.py", line 1147, in run
    yielded = self.gen.send(value)
  File "/srv/conda/lib/python3.6/site-packages/notebook/services/kernels/kernelmanager.py", line 147, in start_kernel
    self.initialize_culler()
  File "/srv/conda/lib/python3.6/site-packages/notebook/services/kernels/kernelmanager.py", line 385, in initialize_culler
    self._culler_callback.start()
  File "/srv/conda/lib/python3.6/site-packages/tornado/ioloop.py", line 1209, in start
    self._schedule_next()
  File "/srv/conda/lib/python3.6/site-packages/tornado/ioloop.py", line 1237, in _schedule_next
    self._update_next(self.io_loop.time())
  File "/srv/conda/lib/python3.6/site-packages/tornado/ioloop.py", line 1244, in _update_next
    callback_time_sec *= 1 + (self.jitter * (random.random() - 0.5))
TypeError: unsupported operand type(s) for *: 'AsyncIOMainLoop' and 'float'
```

</details>

I tried creating this environment locally and had similar problems.  My current guess is that it is related to the `tornado` package version along the lines of https://github.com/ContinuumIO/anaconda-issues/issues/8789  The environment was added in 51f4c91fb6213c80cacaf402d4089b46aa354e65 on 2018-01-24.  conda-forge had `tornado-4.5.3` at that time and upgraded to version 5 shortly thereafter.

@dhimmel are you still able to locally create the `contrib-viz` conda environment?  Or if you have an existing version that works can you please report all of the package versions so we can debug the Jupyter kernel problem?  This will be a general issue for anyone who wants to execute these figures irrespective of binder.

Note also that I had to update the CRAN URL to use the archive URL.


dhimmel:
The kernel not found seems like an error that would occur when nb_conda_kernels does not exist. However, we do install that:

https://github.com/greenelab/meta-review/blob/1bb5c480a22d4ae22dab1541776bdb4f698bc737/analyses/deep-review-contrib/environment.yml#L8

In what kernel or environment are you getting the tornado error?

agitter:
> In what kernel or environment are you getting the tornado error?

I had been getting that error in the `R [conda env:conda]` kernel.  I believe fixing `tornado` in f97f298 helped.  Now in my local environment I can launch the notebook without those errors, but I ran into this dplyr versioning problem https://github.com/r-lib/rlang/issues/674  Upgrading to dplyr 0.7.8 fixed that locally so I changed the environment in eea51b5.  I can run the notebook locally now.

In binder, the kernel error pops up at launch but then I can select the `R [conda env:conda]` kernel and it appears to work.  There are problems running the code though.  The first one is that `readr::read_tsv` doesn't work.  The underlying problem is:
```
OlsonNames()
Warning message in OlsonNames():
“no Olson database found”
```

It's likely related to https://github.com/jupyter/docker-stacks/issues/502

agitter:
Success!  I can now run this notebook in binder: https://mybinder.org/v2/gh/agitter/meta-review/binder?filepath=analyses/deep-review-contrib/02.contrib-viz.ipynb

Installing tzdata with apt-get fixed the biggest remaining issue.  Everything else was conda related.  Upgrading r-dplyr created inconsistencies with most (all?) of the other R packaged.  I ended up upgrading all of them to a combination I found to work locally.  The difficultly in recreating a conda environment that worked a year ago suggests that continuous testing of environments is important if we want to rely on them for notebooks like this, with or without binder.  We may also want to dump `conda list` in addition to `sessionInfo()` to aid environment debugging.

There is still an error message about not finding the matching R kernel.  Selecting `R [conda env:conda]` as the kernel works though.

@dhimmel what do you think?  Should I proceed and add a link to the manuscript and some documentation to the binder build?  I'm not good with ggplot so I'm not going to attempt the dot plot discussed in #149.

agitter:
This has something to do with the kernelspec.  In the notebook it is specified as 

https://github.com/greenelab/meta-review/blob/1bb5c480a22d4ae22dab1541776bdb4f698bc737/analyses/deep-review-contrib/02.contrib-viz.ipynb#L818-L824

The kernel binder recognizes is `R [conda env:conda]`

dhimmel:
> The difficultly in recreating a conda environment that worked a year ago suggests that continuous testing of environments is important if we want to rely on them for notebooks like this, with or without binder.

Yes it's a bit unnerving that despite our efforts to specify the environment, it takes hours to debug issues that have arisen in the last year. It suggests our conda environments do not do a good job of preserving environments across systems and time. It may be that we find more complete environment images are required.

> We may also want to dump conda list in addition to sessionInfo() to aid environment debugging.

In the figure notebook? Yea, not a bad idea.

> Should I proceed and add a link to the manuscript and some documentation to the binder build?

Yeah sounds reasonable.

dhimmel:
> The kernel binder recognizes is R [conda env:conda]

When I download the notebook that is launched via binder (after selecting the R kernel), the .ipynb file includes:

```json
"kernelspec": {
   "display_name": "R [conda env:conda]",
   "language": "R",
   "name": "conda-env-conda-r"
  },
```

Therefore, if we use these details, I think binder will automatically open the proper kernel. Too bad there seems to be an environment naming discrepancy between my local system and binder. I'd say we should use the binder names to make the binder workflow least problematic.

cgreene:
> The difficultly in recreating a conda environment that worked a year ago suggests that continuous testing of environments is important if we want to rely on them for notebooks like this, with or without binder.

Experiences like this are what make me cringe whenever folks suggest that Conda is the answer for what ails us with respect to reproducibility in bioinformatics. 😬

dhimmel:
> Experiences like this are what make me cringe whenever folks suggest that Conda is the answer for what ails us with respect to reproducibility in bioinformatics.

To be fair, Docker only enables reproducibility in a very narrow sense. While there will exist an image that can recreate the output, if you want to make enhancements that require additional software, then you must recreate the image and potentially face issues like we're facing with conda. So in essence, Docker enables exact reproducibility but not necessarily much more, before it experiences the same problems.

cgreene:
@dhimmel : I don't disagree with that either. 😁

cgreene:
I should clarify that I cringe when Docker is suggested as the magic bullet too.

agitter:
I'll proceed to prepare this pull request for review.

> Therefore, if we use these details, I think binder will automatically open the proper kernel.

In df20be8 I tested updating the notebook kernelspec to match what you reported above.  That will at least reveal whether we can get this working in binder without the kernel error.  Then post-merge I'll follow up with the binder team to see what a better solution would be for dynamically fixing the kernelspec.  I don't want to cause this same kernel problem in the local environment.

Given what we've seen with the conda environment, I'll create a new tag for this repository once we have a stable version.  Maybe something like `binder-v1`?  That way binder should always load the cached Docker image from the currently functional environment, if I understand how it works correctly.

dhimmel:
> That way binder should always load the cached Docker image from the currently functional environment, if I understand how it works correctly.

@agitter is there a link with documentation on how binder caches Docker images for a repo? I think ideally we want the repository URL to point to a binder with a cached environment. However, we would like the ability to update this environment in case it breaks, such that the URL in the manuscript does not become stale.

What happens if we just use the latest `master` version for binder? How would binder go about caching the image?

analyses/deep-review-contrib/install-manual-r-dependencies.r
```diff
@@ -1,6 +1,6 @@
 packages = c(
   'https://cran.r-project.org/src/contrib/svglite_1.2.1.tar.gz',
-  'https://cran.r-project.org/src/contrib/ggridges_0.4.1.tar.gz'
+  'https://cran.r-project.org/src/contrib/Archive/ggridges/ggridges_0.4.1.tar.gz'
```
dhimmel: @kurtwheeler is there a URL structure for versioned R dependencies that supports both the latest version and archived versions? IIRC you had to deal with this at some point, and there was an undocumented workaround.

agitter:
@dhimmel this page https://mybinder.readthedocs.io/en/latest/tutorials/reproducibility.html states:
> Binder will create a new Docker image the first time is run with a repository. From then on, a new image will only be created if the commit hash changes (if you’re linking Binder to a branch and not a specific commit hash).

If you'd like more technical information, I may need to open an issue to ask the development team.  I haven't found more details.

You made me realize that my suggestion of making a `binder-v1` tag is a bad idea.  That link would work for the gh-pages version where we can update the URL but not in a published version.

We could use the latest `master`.  The downside is that I expect it to break and require frequent maintenance.  An update to any package in the environment that isn't pinned runs the risk that the environment could become incompatible with the notebook code.  Or as we saw with `tornado` it could break the entire kernel.  There is a risk that users would follow the binder link and arrive at a broken notebook.  I'm not eager to set up another CI service right now that would run cron jobs to confirm the notebooks can still run in the current environment.

What about creating a `binder` branch (or tag?) that always points to the last stable environment?  We would always use the same URL and could decide when to point the branch to a new stable environment as we make updates?

For my own notekeeping and debugging, I also used `jupyter kernelspec list` to dump the kernels that are created:
```
Available kernels:
  ir         /srv/conda/share/jupyter/kernels/ir
  python3    /srv/conda/share/jupyter/kernels/python3
```

dhimmel:
> a new image will only be created if the commit hash changes

It sounds like binder will recreate the docker image whenever new commits are present on a branch. It is not inferring whether the commits are unrelated to the environment and thus skipping the image recreation.

I think a `binder` branch or tag makes sense. Then we would only update this branch/tag when we change `deep-review-contrib`. I think I would prefer a tag, since we will always want `binder` to refer to a commit within the `master` history. Perhaps we can double check that binder will rebuild the image when a tag is changed to refer to another commit (since I think this is perhaps non-standard).

analyses/deep-review-contrib/install-manual-r-dependencies.r
kurtwheeler: There is not such a URL structure. Our solution to this problem is to use [devtools](https://cran.r-project.org/web/packages/devtools/index.html) to install packages rather than specifying URLs directly.

However if you want to use the URL directly [this file](https://github.com/AlexsLemonade/refinebio/blob/b82715bd10aa0d3dbac4455b6e944f337ffdb229/common/install_devtools.R) might be of interest to you.

dhimmel: > There is not such a URL structure

But I remember there being some undocumented way to retrieve it, like with `?version=X.X.X` appended to the URL or something. However, `devtools::install_version` [doesn't seem to use](https://rdrr.io/cran/remotes/src/R/install-version.R) this method I'm thinking of.

Cross-referencing [this tweet](https://twitter.com/BradleyJEck/status/919925180012654592) (unresolved).

slochower:
FYI, I tried to take a look at the version of the notebook on Binder today and it took me maybe 5 minutes to build the image. But it did eventually run. Is this because it doesn't have a cached image yet?

image

analyses/deep-review-contrib/install-manual-r-dependencies.r
dhimmel: It looks like `remotes::install_version` is more lightweight of a dependency than `devtools`.

dhimmel: @kurtwheeler didn't you [get the answer at the mailing list here](http://r.789695.n4.nabble.com/Suggestion-Make-CRAN-source-URLs-immutable-tp4753344.html)? For example, the following URLs should work?

```
https://cran.r-project.org/package=svglite&version=1.2.1
https://cran.r-project.org/package=ggridges&version=0.4.1
```

agitter:
@slochower yes that's possible.  I'm having a hard time debugging the kernel issues.  I've been pushing a few small changes to trigger new builds  so I can watch the build process.  If you happened to load one of those new commits it would be very slow.

That said, even a cached image can take minutes to load.

analyses/deep-review-contrib/install-manual-r-dependencies.r
dhimmel: @agitter can you see if those two URLs work?

slochower:
> That said, even a cached image can take minutes to load.

Got it. I wonder whether this will get shorter over time.

analyses/deep-review-contrib/install-manual-r-dependencies.r
dhimmel: Replied to a stackoverflow with the [undocumented method](https://stackoverflow.com/a/54991606/4651668).

agitter:
It looks like `nb_conda_kernels` doesn't interact well with the conda environment creation process that repo2docker uses as part of binder.  `conda env list` shows
```
# conda environments:
#
base                  *  /srv/conda
```

That is because the environment file is being used to update the existing environment, not create a new environment:
https://github.com/jupyter/repo2docker/blob/e6cbf3e3a3e399508d5628b4ff1d30afed8c944f/repo2docker/buildpacks/conda/__init__.py#L179-L186

analyses/deep-review-contrib/install-manual-r-dependencies.r
agitter: Testing this in 8aff589.  I also changed other parts of the build so I may need to stabilize those before I can tell whether this worked.

dhimmel:
> It looks like nb_conda_kernels doesn't interact well with the conda environment creation process that repo2docker uses as part of binder. conda env list shows

Axe it?

analyses/deep-review-contrib/install-manual-r-dependencies.r
agitter: I watched the build logs and these URLs appeared to work.  As expected, I still need to fix other parts of the build before I can test the packages are functional.

agitter: Installing the R packages with these CRAN URLs worked.

agitter:
It was not easy to create and activate a new conda environment, and I decided to not debug it further.  Now that I understand the build system better, updating the notebook's kernelspec is the best option presented in https://github.com/jupyterhub/binder/issues/152

Removing `nb_conda_kernels` may have worked as well, but this solution allows us to continue using `nb_conda_kernels` for running notebooks in a local environment.

This is functioning and ready for review.  For now it uses the master branch.  We'll need to finalize our tagging strategy before merging or create a new issue to do so post-merge.

analyses/deep-review-contrib/README.md
```diff
@@ -9,3 +9,5 @@ conda env create --file environment.yml
 source activate contrib-viz
 Rscript install-manual-r-dependencies.r
 ```
+
+The environment was updated to be compatible with Binder after the notebooks were originally run.
```
dhimmel, line 13: Do you want to rerun the notebooks and save with the current environment?

content/02.main-text.md
```diff
@@ -97,6 +97,7 @@ The Deep Review was initiated in August 2016, and the first complete manuscript
 While the article was under review, we continued to maintain the project and accepted new contributions.
 The preprint was updated in January 2018, and the article was accepted by the journal in March 2018 [@doi:10.1098/rsif.2017.0387].
 As of June 15, 2018, the Deep Review repository accumulated {{total_commits}} git commits, {{merged_pull_requests}} merged pull requests, {{open_issues + closed_issues}} issues, and {{github_stars}} GitHub stars.
+[Click here](https://mybinder.org/v2/gh/greenelab/meta-review/master?filepath=analyses/deep-review-contrib/02.contrib-viz.ipynb) to launch Binder and interactively explore alternative visualizations in the Jupyter notebook that generated this figure.
```
dhimmel: An alternative phrasing that is not necessarily better than the existing phrasing (so feel free to disregard):

> The notebooks to generate this figure can be [interactively launched using Binder](https://mybinder.org/v2/gh/greenelab/meta-review/master?filepath=analyses/deep-review-contrib/02.contrib-viz.ipynb), enabling users to explore alternative visualizations or analyses of the source data.

binder/postBuild
```diff
@@ -0,0 +1,3 @@
+Rscript analyses/deep-review-contrib/install-manual-r-dependencies.r
+sed -i 's/R \[conda env:contrib-viz\]/R \[conda env:conda\]/g' analyses/deep-review-contrib/02.contrib-viz.ipynb
+sed -i 's/conda-env-contrib-viz-r/conda-env-conda-r/g' analyses/deep-review-contrib/02.contrib-viz.ipynb
```
dhimmel, line 3: I agree this workaround is the best option available.

dhimmel (commented):
I think we probably should update `master` to `binder` (for a binder tag) prior to merging.

agitter:
We have confirmation in https://github.com/jupyterhub/binder/issues/153 that the `binder` tag strategy will work.  @dhimmel should I tag the current `greenelab/meta-review` for now and then update it as soon as this is merged?

binder/postBuild
agitter: 👍

content/02.main-text.md
agitter: I switched to that in ac45b94 but used the singular "notebook".  I haven't tested the Python kernels extensively, and we don't provide a direct link to open them.

I also switched to the binder tag and noted it in 7512bcd.

analyses/deep-review-contrib/README.md
agitter: This was intended as a temporary comment that we can remove after rerunning the notebooks in the new environment.  However, I'm out of the office today and may not trust that the environment works perfectly on this Windows laptop.

I also thought that we might add a dot plot for https://github.com/greenelab/meta-review/issues/149 and then run the notebook one final time after that update.  What do you think?

slochower: > I also thought that we might add a dot plot for #149 and then run the notebook one final time after that update. What do you think?

Yesterday I quickly added `geom_dot()` in place in the figure code, but the dots didn't align correctly with the filled region. `geom_dot()` without the lines worked, but the combination was wonky. I don't have an example to show because I was just messing around in Binder. Perhaps @dhimmel can easily figure out what's up.

agitter: I'm quite bad with R plotting so I'm not planning to make an attempt at this.

slochower: FYI, I just tried to launch Binder using the link at the top of this page, to recreate the figure that I failed to describe with words and Binder failed.

image

slochower: End of logs show...

```
* DONE (ggridges)
Removing intermediate container e636367f983b
 ---> 6d68f48f80dd
Step 47/47 : CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]
 ---> Running in 5785ed11313d
Removing intermediate container 5785ed11313d
 ---> 1bd6c22dcbe3
{"aux": {"ID": "sha256:1bd6c22dcbe378bf58007f2cab2085ccbaf6006e7963fd7726f285c1ae2667ab"}}Successfully built 1bd6c22dcbe3
Successfully tagged gcr.io/binder-prod/r2d-72d7634-agitter-2dmeta-2dreview-39aa7f:7512bcd452a0cbb6dd6241a296b69be9f51475ac
Pushing image
Pushing image
Pushing image
Pushing image
Pushing image
Pushing image
Pushing image
Pushing image
Pushing image
Pushing image
Pushing image
Failed to connect to event stream
```

dhimmel:
>  @dhimmel should I tag the current greenelab/meta-review for now and then update it as soon as this is merged?

Doesn't matter. I might just wait till merge.

analyses/deep-review-contrib/README.md
dhimmel: > I also thought that we might add a dot plot for #149 and then run the notebook one final time after that update. What do you think?

I'm happy to take a crack at this visualization. I suggest we merge this PR first so I can work off of the updated environment.

dhimmel (approved):

analyses/deep-review-contrib/README.md
slochower:
```r
gg = total_words_df %>%
  ggplot2::ggplot(ggplot2::aes(x = date, y = username, height = total_words, group = username, fill=username, color=username)) +
  ggplot2::geom_dotplot(binaxis = "y", dotsize = 0.1) + 
  # ggridges::geom_density_ridges(stat='identity', size=0.4, alpha=0.3) +
  ggplot2::scale_x_datetime(expand = c(0, 0), date_labels = '%b %Y', breaks = major_breaks, minor_breaks = minor_breaks) +
  ggplot2::scale_y_discrete(limits=usernames, position = 'right', labels=y_labels) +
  #ggplot2::scale_fill_manual(values = setNames(usage_df$color, usage_df$username)) +
  ggplot2::guides(fill=FALSE, color=FALSE) +
  ggplot2::xlab(NULL) +
  ggplot2::ylab(NULL) +
  ggplot2::theme_minimal() +
  ggplot2::theme(axis.text.y = ggplot2::element_text(vjust = 0)) +
  ggplot2::theme(panel.grid.major.y = ggplot2::element_blank())
```

This naively looks correct, but dot plot starts around Jul 2017 instead of Jan 2017.

image

Binning in the other direction looks correct, but it's not really the "timeline" view that the reviewer requested.

```r
gg = total_words_df %>%
  ggplot2::ggplot(ggplot2::aes(x = date, y = username, height = total_words, group = username, fill=username, color=username)) +
  ggplot2::geom_dotplot(binaxis = "x", dotsize = 0.2) + 
  ggridges::geom_density_ridges(stat='identity', size=0.4, alpha=0.3) +
  ggplot2::scale_x_datetime(expand = c(0, 0), date_labels = '%b %Y', breaks = major_breaks, minor_breaks = minor_breaks) +
  ggplot2::scale_y_discrete(limits=usernames, position = 'right', labels=y_labels) +
  ggplot2::guides(fill=FALSE, color=FALSE) +
  ggplot2::xlab(NULL) +
  ggplot2::ylab(NULL) +
  ggplot2::theme_minimal() +
  ggplot2::theme(axis.text.y = ggplot2::element_text(vjust = 0)) +
  ggplot2::theme(panel.grid.major.y = ggplot2::element_blank())
```

image

slochower: @dhimmel great! See a few examples above...

agitter: @slochower `Failed to connect to event stream` is not a fatal error in my recent experience.  It indicates an interruption between the build process and the local client that prevents the log from updating in the browser, but the build continues.  Usually refreshing will either reset the build log output (from its current point, not the beginning) or launch the container if it finished building.

agitter:
I'm merging and then will add the `binder` tag.  We can rerun the notebooks with a new visualization and update the readme comment in a new pull request.

analyses/deep-review-contrib/README.md
slochower: Huh. A refresh launched a rebuild, which did work, but it took another 5-10 minutes, so I don't think it simply reset the output. Good to know, though!

PR #163, opened 2019-03-05
Add mention of Binder

betatim:
Seems appropriate as some of the other tools use Binder to do what they do.

The mentions in that sentence seem to be sorted alphabetically so I added Binder at the front but I think it reads a bit weird. Is there a better sorting or place to put it?

Also, do I really have to do nothing else but mention the DOI I want to cite?

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -264,7 +264,7 @@ Manubot provides an ideal platform for perpetual reviews [@arxiv:1502.01329; @ta
 Concepts for the future of scholarly publishing extend beyond collaborative writing [@doi:10.22541/au.149693987.70506124; @tag:techblog-brown].
 Bookdown [@doi:10.1201/9781315204963] and Pandoc Scholar [@doi:10.7717/peerj-cs.112] both extend traditional Markdown to better support publishing.
 Examples of continuous integration to automate manuscript generation include [gh-publisher](https://github.com/ewanmellor/gh-publisher) and Continuous Publishing [@url:http://blog.martinfenner.org/2014/03/10/continuous-publishing/], which was used to produce the book Opening Science [@doi:10.1007/978-3-319-00026-8].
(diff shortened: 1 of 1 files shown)


agitter:
> Is there a better sorting or place to put it?

I agree it is appropriate to mention and cite Binder here.  I'm wondering whether it should be a separate sentence though.  Is Binder an interactive "manuscript" per se or a more general relevant technology that can support interactive manuscripts among other things?

Note that we also did find and cite that Binder 2.0 reference in our Figure 2 caption.  Adding it here again is still relevant though.

> Also, do I really have to do nothing else but mention the DOI I want to cite?

Yes!  That's one of Manubot's most convenient features.  It supports citation by identifier for a variety of [standard identifiers](https://github.com/manubot/rootstock/blob/master/USAGE.md#citations).

betatim:
>  Is Binder an interactive "manuscript" per se or a more general relevant technology that can support interactive manuscripts among other things?

I think it is in the latter category. There are a few tools built on top of Binder like [jupyter-book](https://jupyter.org/jupyter-book/03/1/Expressions.html) and [juniper](https://spacy.io/usage/linguistic-features) (scroll down to the code snippet and click run) which show off the "published manuscript with interactive code". So maybe right now Binder and tools around it are in the category "toolkit for you to build your own interactive manuscript solution".

It would be interesting (because I don't know how to do it) to see how how advanced the Stencila tooling is for turning https://github.com/stencila/examples/tree/master/elife-30274 (which I think is the "source") into https://repro.elifesciences.org/example.html# (which is the "manuscript" and links to https://github.com/elifesciences/rds-example as its source/what is being served). If we have this we could make a judgement relative to Stencila if the Binder "know which tools you need to combine" approach is still too rough/prototype stage.

Either way I think it remains true that Binder is in the category "enablers" or "foundation for lots of other tools" (Stencila, jupyter book, juniper, ...) rather than being a manuscript tool itself.

agitter:
After re-reading this, the original phrasing may be okay and I don't want to over-complicate this paragraph.  It is legitimate to say that Binder "supports" this type of manuscript.

I'll merge and we can revisit if needed.

agitter (approved):

PR #164, opened 2019-03-06
Update deep-review-contrib: environment & dot plot

dhimmel:
This pull requests follows up on https://github.com/greenelab/meta-review/pull/161#discussion_r262663346 and updates the environment for deep-review-contrib. Since we're going to be adding a new figure, I though I might as well modernize the environment a bit.

@agitter do you see anything here that will interfere with binder. I haven't tested it with binder.


dhimmel:
In https://github.com/greenelab/meta-review/pull/164/commits/60bb5243a2c18cf5e67a22181fefac43de313149, I add a dot plot. The design was inspired by reviewer comments in https://github.com/greenelab/meta-review/issues/149. See below:



One nice aspect is that we can show additions and deletions. The outer circle indicates the category with a greater word count. For example, a green outline with red circle indicates the changes within a month included more additions than deletions.

agitter:
I'm reviewing the changes now.  Regarding Binder, we can test any public branch or commit by changing the Binder URL.  In this case https://mybinder.org/v2/gh/dhimmel/meta-review/dot-plot?filepath=analyses/deep-review-contrib/02.contrib-viz.ipynb should work.

analyses/deep-review-contrib/environment.yml
```diff
@@ -2,23 +2,25 @@ name: contrib-viz
...
+- conda-forge::python=3.6.7
+- conda-forge::pyyaml=3.13
+- conda-forge::requests=2.21.0
+- conda-forge::tornado=4.5.3
 - conda-forge::r-base=3.5.1
 - conda-forge::r-cowplot=0.9.4
 - conda-forge::r-dplyr=0.7.8
 - conda-forge::r-gdtools=0.1.7
 - conda-forge::r-ggplot2=3.1.0
+- conda-forge::r-ggridges=0.5.1
```
agitter, line 21: That certainly simplifies things.  Discovering the new type of CRAN URL was still a valuable exercise for future use cases though.

dhimmel:
> In this case https://mybinder.org/v2/gh/dhimmel/meta-review/dot-plot?filepath=analyses/deep-review-contrib/02.contrib-viz.ipynb should work.

I loaded the binder using that link, and got the dreaded:

image

There are two R kernels available for selection, but both of them seem to load indefinitely (busy circle filled in, with a No Kernel message).

binder/postBuild
```diff
@@ -1,3 +1,2 @@
-Rscript analyses/deep-review-contrib/install-manual-r-dependencies.r
 sed -i 's/R \[conda env:contrib-viz\]/R \[conda env:conda\]/g' analyses/deep-review-contrib/02.contrib-viz.ipynb
```
agitter: One of the environment updates changed the kernel behavior in Binder.  I get the kernel not found error again.  Downloading the notebook from Binder shows
```
 "metadata": {
  "anaconda-cloud": {},
  "kernelspec": {
   "display_name": "R [conda env:root] *",
   "language": "R",
   "name": "conda-root-r"
  },
```

Can you please try updating the display name and name in these `sed` commands?

slochower:
And I loaded the Binder using that link and got Binder inaccessible.

image

Edit: hit refresh and got @dhimmel 's Kernet not found error message.

agitter (changes requested):
Excellent updates.  The Binder notebook works once I manually set the kernel, which we should be able to automate by modifying the `sed` lines.

The visual style of the dot plot is fantastic, I like how it turned out.  I was a little surprised by how many deletions there are.  The cumulative deletions almost seems to match the cumulative additions, which is counter-intuitive because the overall length grew over time.  Is this because a one word edit is reflected as both an addition and a deletion?  For some contributors who make small contributions like Aaron Sheldon, it appears as if they only deleted text.

I don't have suggestions for the dot plot, and we can decide which figure to use in the manuscript in #149 after merging this.  We may want to emphasize in the caption that we have three alternative visualizations in the notebook.

I also noticed a typo in cell 7 of `01.contrib-git.ipynb`: `begore renaming`

agitter:
@dhimmel that should be easy to fix, see my review comments I just added.  I believe it is due to the `nb_conda_kernels` update.

@slochower I got that once as well for this notebook.  Reloading the page fixed it.

slochower:
> @slochower I got that once as well for this notebook. Reloading the page fixed it.

Yup, a reload fixed the problem, in the sense that I was then able to get the kernel not found error. Although you have done almost all of the heavy lifting to get Binder working, I must say that even _I_ feel like this is not as straightforward and robust as I expected & hoped.

analyses/deep-review-contrib/environment.yml
dhimmel: Yes, I'm glad we have that in the commit history.

agitter (approved):

dhimmel:
As of https://github.com/greenelab/meta-review/commit/38f865ae500298867825b47dfb08f172237d82fb, I was able to rerun the `02.contrib-viz.ipynb` on binder.

dhimmel:
> I was a little surprised by how many deletions there are. The cumulative deletions almost seems to match the cumulative additions, which is counter-intuitive because the overall length grew over time. Is this because a one word edit is reflected as both an addition and a deletion? For some contributors who make small contributions like Aaron Sheldon, it appears as if they only deleted text.

It's important to think of these as deletions and additions in the git sense. For example, if you edit one word that counts as an addition and deletion of both 1. I plotted the smaller circle on top, so oftentime a dot will be mostly red with a green ring, indicating more words were added than deleted. I don't love that the dots don't show the entire area of deletions and additions (since the points are plotted on top of each other), but it's a hard problem to solve.

Issue #126, opened 2019-02-07
Update acknowledgements


agitter:
We should review the manubot and manubot-rootstock issues and updates to see who else should be acknowledged in the revised manuscript.  In addition, we should acknowledge the public reviewer.

agitter:
I propose we also ask John MacFarlane if we can acknowledge him.  He has been very helpful with pandoc usage or debugging.

dhimmel:
> I propose we also ask John MacFarlane if we can acknowledge him. He has been very helpful with pandoc usage or debugging.

Agreed. @jgm would you be okay with us acknowledging you in our manuscript describing the Manubot tool? We'd add it to [this section](https://greenelab.github.io/meta-review/v/c4baa224a99d86d88e32c2dab67245c47838fbde/#acknowledgments). Manubot uses pandoc along with other open source tools to create automated workflows for scholarly publishing. Thanks for your help across many pandoc issues over the last year or so!

jgm:
Sure, no objections!

agitter:
As a reminder to myself, we should also remove Dongbo from the acknowledgements when we add him as an author.

agitter:
@zach-hensel we're revising a manuscript describing Manubot and collaborative writing.  Can we add you to the acknowledgments for your help with Manubot testing and debugging?

agitter:
I can update the text when it is time. @dhimmel and @vincerubinetti can you please help me recall any other new manubot or rootstock contributors?

dhimmel:
> can you please help me recall any other new manubot or rootstock contributors?

@lierdakil has provided helpful advice in several discussions such as https://github.com/manubot/rootstock/pull/196#issuecomment-475473699 and https://github.com/lierdakil/pandoc-crossref/issues/30#issuecomment-124023824. @lierdakil want to give this comment a thumbs up if you are okay with us acknowledging you in our [manuscript](https://greenelab.github.io/meta-review/v/d4a642efa06896209ddfa6e3d2387a58232e61f5/)?

vincerubinetti:
Not sure the level you want to go with the acknowledgments, but a bunch of people and the Greene Lab helped me by giving feedback on the logo and front-end. We've also had a few people point out small bugs.

agitter:
@vincerubinetti we tend to be very inclusive with acknowledgements, especially for this manuscript.  If you think of anyone, please feel free to tag them here so they can indicate their approval for being included, as @dhimmel and I did above.

[...]

PR #165, opened 2019-03-07
Add pandoc and Binder acknowledgements

agitter:
John approved in #126.  @betatim is it okay with you if you acknowledge you here?


dhimmel (approved):

betatim:
Sure thing. Even better would be if we could find a way to say "thanks to the Binder community" or some such. Binder is a team sport.

How about: "and the global Binder team for advice on Binder."? It is a bit clunky in terms of English language but I think better captures the spirit.

agitter:
> How about: "and the global Binder team for advice on Binder."?

Sure, changed in 04b86ce.

agitter:
@dhimmel does this still look okay with you?

dhimmel (approved):

PR #180, opened 2019-03-25
Add revised workflow figures

cgreene:
This adds the new workflow figures. I'll also update the legend.

Issue #147, opened 2019-02-08
Reviewer 2 comment 18


agitter:
> Table 1. → You should add a WYSIWIG feature that is present in a number of platforms and important for a lot of non technical people. Same for inline comments and diff colorization.

vincerubinetti:
I'd agree with this. Perhaps it could go in "Limitations" as well.

Would a mention of the Hypothesis annotations plugin somewhere make sense? Would that address "inline comments"? Also could you elaborate on "diff colorization"?

agitter:
What are our thoughts about whether Manubot has WYSIWIG support?  It is possible to preview the edited markdown and render elements like lists and formatting.  However, the figure references, citations, and other aspects are not WYSIWIG.

I'm not sure about inline comments.  It could refer to the other platforms and which of them support commenting on specific lines.  Most of them do.  In general, this entire table may need a refresh.  We should switch to Overleaf v2 now that v1 is no longer available (#27).

We should review our text to make sure the inline commenting on Manubot manuscripts through Hypothesis is clearly stated.

For diff colorization, my interpretation is that this would annotate whether the writing platform can diff text updates and color the changes, similar to the GitHub pull request web interface.

slochower:
> What are our thoughts about whether Manubot has WYSIWIG support? It is possible to preview the edited markdown and render elements like lists and formatting.

Is it worth stating that many editors like Atom, VS Code, Sublime... can do interactive preview of Markdown documents? On the one hand, that is not really Manubot-specific in any way, but on the other hand, if not everyone knows this and they are considering using Manubot, it could be helpful.

> We should review our text to make sure the inline commenting on Manubot manuscripts through Hypothesis is clearly stated.

Does anyone use Hypothesis to comment during the drafting process? Do comments in Hypothesis get preserved across pushes to `gh-pages`?

dhimmel:
> What are our thoughts about whether Manubot has WYSIWIG support?

There is no WYSIWIG editor and it is not on the roadmap. Of course, you can use a partial WYSIWIG solution or a preview solution, but it won't understand the full array of formatting options. If you need WYSIWIG, something like [Manuscripts App 2.0](https://www.manuscriptsapp.com) (not released yet) or another browser based editor probably makes the most sense. I think this is a limitation of Manubot for the time being, but one of the reasons we have a specific target audience in mind that is familiar with writing source text files.

> Same for inline comments and diff colorization

[...]

PR #182, opened 2019-03-27
Review 3 subscript and workflow caption clarifications

### LINKED ISSUE #116: footnote placement in rendered manuscript should be left-adjusted
in the PDF I have in front of me, the footnotes at end of sentences are not left-flush with the period. this is a minor nit but a nit nonetheless :). is this just a stylesheet issue or ...?

### LINKED ISSUE #119: "it is reviewed by a maintainer and a participant"
are these separate people or the same person? one presumes maintainers are participants, but not vice versa. either way, please clarify! :)

agitter:
Closes #116
Closes #119 

The figure caption change will create a conflict with #180, which I can resolve after we merge #180.

I'm not sure what the best formatting is for the table subscripts in #116.  As far as I know, we don't directly support footnotes.  It would be too hard to find them if we place footnotes at the end of an HTML document.  Placing them in the table caption is confusing because they were not separated by line breaks.  I attempted an ordered list to add the line breaks, but this formatting is still not quite what I wanted.

Any suggestions?  Should we switch to a more complex HTML-based formatting for this caption?


dhimmel (approved):

agitter:
@dhimmel the list did not render as a list in the Table caption

image

This is not exactly what I had intended for the new formatting, but it is still better than what we had before.  Do you think we need to find a way to add newlines after each numbered footnote?

dhimmel:
I don't think a numbered list within the caption is a great idea. It would be nice if it worked everywhere, but I think there is a high chance at least one place (like the journal) will not support it.

What about doing "1—Additional functionality" or "1) Additional functionality"

agitter:
I like 1) Additional functionality".  I'll make that edit when I updated the table for #147.

agitter:
I updated the caption in 6e2d476 in #188.

Issue #114, opened 2019-02-05
introductory paragraph needs more citations.


ctb:
as is, the introductory paragraph is missing citations for statements.

in particular, the statement "However, the scientific community requires tools whose workflows encourage openness" should either have a citation or two, or be clarified as an opinion.

agitter:
@dhimmel do you know of references we should use here?  If not, I'll help look for them.

dhimmel:
@agitter I forgot to update response-to-reviewers. Can you take a look at https://github.com/dhimmel/meta-review/commit/7e489f99e7a2e1ad76886bab78f3b4c2f672a80f?

PR #189, opened 2019-04-04
Introductory cites and other citation tweaks

dhimmel:
Closes https://github.com/greenelab/meta-review/issues/114

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -341,7 +341,7 @@ Although these criteria provided a straightforward, equitable way to determine w
 In biomedical journals, the convention is that the first and last authors made the most substantial contributions to the manuscript.
 This convention can be difficult to reconcile in a collaborative effort.
 Using Git, we could quantify the number of commits each author made or the number of sentences an author wrote or edited, but these metrics discount intellectual contributions such as discussing primary literature and reviewing pull requests.
-Therefore, we concluded that it is not possible to construct an objective system to compare and weight the different types of contributions and produce an ordered author list.
+Therefore, we concluded that it is not possible to construct an objective system to compare and weight the different types of contributions and produce an ordered author list [@url:http://ivory.idyll.org/blog/2019-authorship-revisiting.html].
 
 To address this issue, we generalized the concept of "co-first" authorship, in which two or more authors are denoted as making equal contributions to a paper.
 We defined four types of contributions [@doi:10.1098/rsif.2017.0387], from major to minor, and reviewed the GitHub discussions and commits to assign authors to these categories.
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -203,7 +203,7 @@ The architecture of the plugins and themes is designed to provide authors with e
 
 ![
 **Examples of the various Manubot plugins, illustrating their functionality and usefulness.**
-Screenshots were taken from existing manuscripts made with Manubot: [Sci-Hub Coverage Study](https://greenelab.github.io/scihub-manuscript/v/fd7acb7ed0108c920da56f84819ce13f02f68aa8/) and [TPOT-DS](https://trang1618.github.io/tpot-ds-ms/), available under the [CC BY 4.0 License](https://creativecommons.org/licenses/by/4.0/ "Creative Commons Attribution 4.0 International License").
+Screenshots were taken from existing manuscripts made with Manubot: [Sci-Hub Coverage Study](https://greenelab.github.io/scihub-manuscript/v/fd7acb7ed0108c920da56f84819ce13f02f68aa8/) and [TPOT-FSS](https://trang1618.github.io/tpot-fss-ms/), available under the [CC BY 4.0 License](https://creativecommons.org/licenses/by/4.0/ "Creative Commons Attribution 4.0 International License").
 Clarifying markups are overlaid in purple.
 ](images/plugins.png "Manubot Plugins"){#fig:plugins}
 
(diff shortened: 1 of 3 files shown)


dhimmel:
The TPOT-DS manuscript is changing names to TPOT-FSS ("dataset selector" to "feature set selector"). In https://github.com/greenelab/meta-review/pull/189/commits/14da1c91c7e30995d4f0e4b99db3dd4a5c2f33e1, I update the links on the meta-review end. However, the repo relocation means the old hypothesis annotations have not transferred. There may be a [workaround](https://web.hypothes.is/help/the-web-page-i-annotated-has-moved-to-a-different-url-and-my-annotations-are-gone-what-can-i-do/) that the TPOT-FSS manuscript can implement:

> Publishers: If you add a tag the new page that refers to the original one, Hypothesis will display annotations on both pages.

I will work with @trang to see what we can do. This can happen independently of this PR.

content/02.main-text.md
```diff
@@ -2,7 +2,7 @@
 
 The internet enables science to be shared in real-time at a low cost to a global audience.
 This development has decreased the barriers to making science open, while supporting new massively collaborative models of research.
```
agitter: I looked for references to support `This development has decreased the barriers to making science open`.  Here's what I have so far:

https://doi.org/10.1038/s41562-016-0021
> For most of the history of scientific publishing, two functions have been confounded — evaluation and dissemination. Journals have provided dissemination via sorting and delivering content to the research community, and gatekeeping via peer review to determine what is worth disseminating. However, with the advent of the internet, individual researchers are no longer dependent on publishers to bind, print and mail their research to subscribers. Dissemination is now easy and can be controlled by researchers themselves...

http://ivory.idyll.org/blog/2016-what-is-open-science.html discusses how senior scientists used to a pre-internet era are a challenge to open science
>  In particular, most senior scientists entered science at a time when the Internet was young and the basic modalities and culture of communicating and sharing over the Internet hadn’t yet been developed.

content/02.main-text.md
```diff
@@ -2,7 +2,7 @@
...
-However, the scientific community requires tools whose workflows encourage openness.
+However, the scientific community requires tools whose workflows encourage openness [@doi:10.17226/25116].
 Manuscripts are the cornerstone of scholarly communication, but drafting and publishing manuscripts has traditionally relied on proprietary or offline tools that do not support _open scholarly writing_, in which anyone is able to contribute and the contribution history is preserved and public.
```
agitter, line 6: Do we need support for the claim `drafting and publishing manuscripts has traditionally relied on proprietary or offline tools that do not support _open scholarly writing_`?  It seems self-evident, but https://doi.org/10.1038/514127a (which we already cite) could be relevant.
> Writing a paper is traditionally a stepwise process. One author shares drafts with colleagues and then waits for everyone to reply or moves forward independently, folding in revisions and queries as they arrive.

> A minority of researchers use these apps [_shareLaTeX, Fidus Writer and Authorea_]

content/02.main-text.md
```diff
@@ -203,7 +203,7 @@ The architecture of the plugins and themes is designed to provide authors with e
 
 ![
 **Examples of the various Manubot plugins, illustrating their functionality and usefulness.**
-Screenshots were taken from existing manuscripts made with Manubot: [Sci-Hub Coverage Study](https://greenelab.github.io/scihub-manuscript/v/fd7acb7ed0108c920da56f84819ce13f02f68aa8/) and [TPOT-DS](https://trang1618.github.io/tpot-ds-ms/), available under the [CC BY 4.0 License](https://creativecommons.org/licenses/by/4.0/ "Creative Commons Attribution 4.0 International License").
+Screenshots were taken from existing manuscripts made with Manubot: [Sci-Hub Coverage Study](https://greenelab.github.io/scihub-manuscript/v/fd7acb7ed0108c920da56f84819ce13f02f68aa8/) and [TPOT-FSS](https://trang1618.github.io/tpot-fss-ms/), available under the [CC BY 4.0 License](https://creativecommons.org/licenses/by/4.0/ "Creative Commons Attribution 4.0 International License").
```
agitter, line 206: Should we do something for the TPOT-DS link in the Figure 3 caption in this pull request?  That link is currently broken.

agitter (changes requested):
Everything you added and changed looks good.  I have a few more ideas that you can consider to see whether they are appropriate.

content/02.main-text.md
dhimmel: Yes this line that you commented on changes it to https://trang1618.github.io/tpot-fss-ms/, which currently works.

agitter: Sorry, not sure what I was thinking.  I thought this was the reference in the example manuscript section, which you also changed.

content/02.main-text.md
dhimmel: Regarding the reviewer comment in https://github.com/greenelab/meta-review/issues/114

> "However, the scientific community requires tools whose workflows encourage openness" should either have a citation or two, or be clarified as an opinion.

I take a different approach. All statements are opinions. Statements with citations are opinions with potential third-party support. It is not necessary to state "in our opinion" because everything is an opinion and readers can choose to agree or disagree with any statement.

That being said, I think intro citations are helpful to direct readers to related work and additional background information. With the current manuscript we happen to include much of this "introductory" stuff in the discussion. I don't think we should add citations unless they are highly appropriate.

We could add many citations to sentences in this paragraph, but if they're just referencing one paragraph of opinion from another study, what's the point? So in summary, unless a citation provides quantitative evidence or a large extent of discussion on a topic, I don't see a need to include it.

I'll also look for a few additional citations, but in general want to prevent low quality citations sneaking in right before resubmission to satisfy a reviewer, which is a common problem I see. @agitter not saying your suggested citations are bad. I will also take a closer look at those.

dhimmel: This was before my time, but I think the [E-Biomed proposal](https://profiles.nlm.nih.gov/ps/access/MVBBWN.pdf) was motivated by decreased technological barriers to openness. However, this proposal never overcame the political barriers as discussed in "[The Remarkable Transformation of E-Biomed into PubMed Central](https://scholarworks.iu.edu/dspace/bitstream/handle/2022/170/wp01-03B.html)"

dhimmel: [This revolution will be digitized: online tools for radical collaboration](https://doi.org/10.1242/dmm.003285) is a possible citation for:

> This development has decreased the barriers to making science open, while supporting new massively collaborative models of research.

content/02.main-text.md
```diff
@@ -341,7 +341,7 @@ Although these criteria provided a straightforward, equitable way to determine w
 In biomedical journals, the convention is that the first and last authors made the most substantial contributions to the manuscript.
 This convention can be difficult to reconcile in a collaborative effort.
 Using Git, we could quantify the number of commits each author made or the number of sentences an author wrote or edited, but these metrics discount intellectual contributions such as discussing primary literature and reviewing pull requests.
-Therefore, we concluded that it is not possible to construct an objective system to compare and weight the different types of contributions and produce an ordered author list.
+Therefore, we concluded that it is not possible to construct an objective system to compare and weight the different types of contributions and produce an ordered author list [@url:http://ivory.idyll.org/blog/2019-authorship-revisiting.html].
```
dhimmel, line 344: @agitter I think this citation fits here and should be included somewhere in the manuscript because it's about authorship for open collaborative projects / manuscripts

content/02.main-text.md
agitter: The references I listed above are third-party opinion, so I see your point about not needing to reference others' opinion to support our own.

agitter (approved):
I'm approving because I have no further suggestions.  If you find any other citations you want to add, I can take another look.

dhimmel:
@agitter can you re-review with https://github.com/greenelab/meta-review/pull/189/commits/a9a5a347dd8de195470cba7c5c585375ffbdc87e. I commented this somewhere else in the wrong location...

agitter (approved):
Good catch.  I approve the response to reviewers as well.

PR #192, opened 2019-04-06
Add description of private manuscripts

### LINKED ISSUE #138: Reviewer 2 comment 9
> Is it possible to use manubot on a private GitHub repository? I'm pretty sure some people would be interested in using manubot but probably would not use it for open editing.

agitter:
Closes #138

WIP until the discussion in #138 with @vsmalladi is resolved.

I'm imaging a workflow where the user has a private repository and disables GitHub Pages without modifying the build script or anything else.  In that case, they could obtain the HTML and PDF from the `gh-pages` branch.  They could view the PDF on GitHub but not the HTML, which would have to be viewed locally.

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -227,7 +227,8 @@ Furthermore, versioned URLs, such as <https://greenelab.github.io/meta-review/v/
 
 The Manubot examples in this manuscript use open public GitHub repositories.
 However, it is also possible to take advantage of Manubot's continuous publishing for a private manuscript by using a private GitHub repository and disabling GitHub Pages.
-In this private setting, the manuscript HTML and PDF outputs are automatically generated when the manuscript source is updated, but only GitHub users granted access to the repository are able to download them.
+This requires additional custom configuration.
+In the private setting, the manuscript HTML and PDF outputs are automatically generated when the manuscript source is updated, but only GitHub users granted access to the repository are able to download them.
 
 ### Timestamping
 
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -225,6 +225,10 @@ If successful, the output is deployed back to GitHub (to dedicated [`output`](ht
 As a result, <https://greenelab.github.io/meta-review> stays up to date with the latest HTML manuscript.
 Furthermore, versioned URLs, such as <https://greenelab.github.io/meta-review/v/4b6396bcefd1b9c7ddf39c1d3f0b3eab2dd63f31/>, provide access to previous manuscript versions.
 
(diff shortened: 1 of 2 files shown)


agitter:
This is now ready for review.  We need to finalize the instructions we provide users, but we have sufficient evidence that this will work.

vsmalladi (approved):
Looks good to me. I have proved that it can also work in private gitlab but having issues with making publishing the pages.

content/02.main-text.md
```diff
@@ -225,6 +225,11 @@ If successful, the output is deployed back to GitHub (to dedicated [`output`](ht
 As a result, <https://greenelab.github.io/meta-review> stays up to date with the latest HTML manuscript.
 Furthermore, versioned URLs, such as <https://greenelab.github.io/meta-review/v/4b6396bcefd1b9c7ddf39c1d3f0b3eab2dd63f31/>, provide access to previous manuscript versions.
 
+The Manubot examples in this manuscript use open public GitHub repositories.
+However, it is also possible to take advantage of Manubot's continuous publishing for a private manuscript by using a private GitHub repository and disabling GitHub Pages.
```
dhimmel: With GitHub Pages disabled, it's odd to call it "continuous publishing"... It is sort of because the outputs are still there as noted below.

In the private repo use case, I imagine a large portion of users will actually keep GH Pages enabled while keeping the URL private.

dhimmel (commented):
I think the proposed paragraph is well-written but have reservations about including it in the manuscript. Especially with the lack of real-world private manubot repo usage. Basically, our understanding of private Manubot repos is a bit speculative at this point.

It seems that the current text would fit well in response-to-reviewers and that we should add a section to Rootstock's `SETUP.md` that provides more detailed instructions on private repos. I can make that Rootstock PR.

I think we should add a sentence to the manuscript like:

> It is also possible to use Manubot on a private GitHub repository, although this can require some additional customization to disable GitHub Pages and may require a paid continuous integration plan.

It would fit well around perhaps:

```
In cases where outside contributions are unwanted, authors can disable pull requests on GitHub.
See the [existing manuscripts] for examples of the range of contribution workflows and Manubot use cases.
```

@agitter what do you think?

agitter:
> Basically, our understanding of private Manubot repos is a bit speculative at this point.

Yes, I have some concerns about that as well.  I do want to point out that this option exists, and your proposed new paragraph does that.  I'll make the updates.

agitter:
@dhimmel ready for re-review.

dhimmel (approved):

PR #194, opened 2019-04-06
Address a non-technical audience

### LINKED ISSUE #130: Reviewer 2 comment 1
> My main concern (+ comments and questions below) is that the article is a bit technical and as such, addresses a technical audience. Since the paper has been submitted as a Software paper, maybe it's fine but it would be nice if non-technical people would also be able to read the paper (to realize open-source collaborative frameworks exists).

agitter:
Closes #130 

@dhimmel I had time to start on this.  I'm happy to revise the intro with you.  My goal is to introduce the main reasons someone would want to use Manubot with non-technical language.  I'll also go through the rest of the manuscript to see if I can explain any overly technical parts.

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -72,11 +72,11 @@ In a primary research article, issues can instead track progress on specific fig
 Issues serve as an open to-do list and a forum for debating the main messages of the manuscript.
 
 GitHub and the underlying Git version control system [@doi:10.1371/journal.pcbi.1004668; @doi:10.1371/journal.pcbi.1004947] also structure the writing process.
-The official version of the manuscript is _forked_ by individual contributors.
+The official version of the manuscript is _forked_ by individual contributors, creating a copy they can freely modify.
 A contributor then adds and revises files, grouping these changes into _commits_.
 When the changes are ready to be reviewed, the series of commits are submitted as a _pull request_ through GitHub, which notifies other authors of the pending changes.
-GitHub's review interface allows anyone to comment on the changes, globally or at specific lines, asking questions or requesting modifications, as described by Johnny Israeli [@url:https://medium.com/towards-data-science/opportunities-and-obstacles-for-deep-learning-in-biology-and-medicine-6ec914fe18c2].
-Conversations during review can reference other pull requests, issues, or authors, linking the relevant people and content, as illustrated in Figure @fig:workflow.
+GitHub's review interface allows anyone to comment on the changes, globally or at specific lines, asking questions or requesting modifications [@url:https://medium.com/towards-data-science/opportunities-and-obstacles-for-deep-learning-in-biology-and-medicine-6ec914fe18c2].
+Conversations during review can reference other pull requests, issues, or authors, linking the relevant people and content (Figure @fig:workflow).
 Reviewing batches of revisions that focus on a single theme is more efficient than independently discussing isolated comments and edits and helps maintain consistent content and tone across different authors and reviewers.
 Once all requested modifications are made, the manuscript maintainers, a subset of authors with elevated GitHub permissions, formally approve the pull request and merge the changes into the official version.
 The process of writing and revising material can be orchestrated through GitHub with a web browser or a local text editor.
@@ -126,7 +126,7 @@ For each manuscript, there is a corresponding Git repository.
 The `master` branch of the repository contains all of the necessary inputs to build the manuscript.
 Specifically, a `content` directory contains one or more Markdown files that define the body of the manuscript as well as a metadata file to set information such as the title, authors, keywords, and language.
 Figures can be hosted in the `content/images` subdirectory or elsewhere and specified by URL.
-Repositories contain scripts and other assets that define how to build and deploy the manuscript.
+Repositories contain scripts and other files that define how to build and deploy the manuscript.
 Many of these operations are delegated to the `manubot` [Python package](https://github.com/manubot/manubot) or other dependencies such as Pandoc, which converts between document formats, and Travis CI, which builds the manuscript in the cloud.
 Manubot pieces together many existing standards and technologies to encapsulate a manuscript in a repository and automatically generate outputs.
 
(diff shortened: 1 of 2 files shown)


agitter:
`REST API` is not defined in Table 2, but there isn't room in the table to explain these two terms.  We should do something to address this.

These sentences are also fairly technical:
> However, sometimes external resources provide Manubot with invalid CSL Data, which can cause errors with downstream citation processors, such as [pandoc-citeproc](http://hackage.haskell.org/package/pandoc-citeproc). Therefore, Manubot removes invalid fields according to the [CSL Data specification](https://github.com/citation-style-language/schema).

Will we lose anything if we remove them and add this to the Manubot documentation instead?  We would need to move the `pandoc-citeproc` reference elsewhere.

content/02.main-text.md
```diff
@@ -19,6 +19,11 @@ The Deep Review and other studies that subsequently adopted the Manubot platform
 However, inviting wide authorship brought many technical and social challenges such as how to fairly distribute credit, coordinate the scientific content, and collaboratively manage extensive reference lists.
 The manuscript writing process we developed using the Markdown language, the GitHub platform, and our new Manubot tool for automating manuscript generation addresses these challenges.
 
+Manubot supports citations by adding a standard identifier like a Digital Object Identifier (DOI) or PubMed Identifier (PMID) directly in the text so that large groups of authors do not have to coordinate reference lists.
+When text is changed, Manubot automatically updates the manuscript's web page so that all authors can read and edit from the latest version.
+Because Manubot manuscripts are created from GitHub repositories, all edits are reviewed and discussed, ensuring that the collaborative text has a cohesive style and message and that authors receive precise credit for their work.
```
dhimmel: > Because Manubot manuscripts are created from GitHub repositories, all edits are reviewed and discussed

What about users who commit directly to the master branch and don't use PRs? IMO Manubot allows a lot of flexibility in terms of how commits are added to the `master` branch. For some projects, it makes sense to use PRs. For others, it may not.

content/02.main-text.md
```diff
@@ -132,30 +137,30 @@ The [Markdown standard](https://spec.commonmark.org/0.28/) itself provides limit
...
+In cases where automatic retrieval of metadata fails or produces incorrect references --- which is most common for URL citations --- users can manually provide the correct CSL JSON using common reference formats.
 Manual CSL JSON also supports references without standard identifiers, such as print-only newspaper articles.
 
 | Identifier | Metadata source | Example citation | Processed citation |
 | --- | --- | --- | --- |
-| Digital Object Identifier (DOI) | DOI [Content Negotiation](https://citation.crosscite.org/docs.html) | `doi:10.1098/rsif.2017.0387` | [@doi:10.1098/rsif.2017.0387] |
+| DOI | DOI [Content Negotiation](https://citation.crosscite.org/docs.html) | `doi:10.1098/rsif.2017.0387` | [@doi:10.1098/rsif.2017.0387] |
 | shortDOI | DOI Proxy Server REST API | `doi:10/gddkhn` | [@doi:10/gddkhn] |
-| PubMed Identifier (PMID) | NCBI [E-utilities](https://www.ncbi.nlm.nih.gov/books/NBK25501/) | `pmid:25851694` | [@pmid:25851694] |
+| PMID | NCBI [E-utilities](https://www.ncbi.nlm.nih.gov/books/NBK25501/) | `pmid:25851694` | [@pmid:25851694] |
```
dhimmel: I don't think it hurts to define acronyms twice, especially in the context of this table.

content/02.main-text.md
```diff
@@ -242,10 +246,12 @@ We find that this asynchronous design with timestamps precise to the day is suit
...
+Fortunately, lock-in to these services is minimal, and [several substitutes](#future) already exist.
+Manubot provides a substantial step towards end-to-end document reproducibility, where every figure or piece of data in a manuscript can be traced back to its origin [@url:https://elifesciences.org/for-the-press/e6038800/elife-supports-development-of-open-technology-stack-for-publishing-reproducible-manuscripts-online] and is well-suited for preserving provenance.
 For example, figures can be specified using versioned URLs that refer to the code that created them.
 In addition, manuscripts can be templated, so that numerical values or tables are inserted directly from the repository that created them.
+Figure @fig:contrib provides examples of templates.
+Phrases such as "{{total_commits}} Git commits" are written as `{{total_commits}} Git commits` so that the commit count can be automatically updated.
```
dhimmel: I think you'll need to do:

```
{% raw %}`{{total_commits}} Git commits`{% endraw %}
```

http://jinja.pocoo.org/docs/2.10/templates/#escaping

dhimmel:
> REST API is not defined in Table 2, but there isn't room in the table to explain these two terms. We should do something to address this.

Currently the table cell is "DOI Proxy Server REST API", which I used because that is was the DOI factsheet [used](http://www.doi.org/factsheets/DOIProxy.html#rest-api):

> The DOI system Proxy Server REST API allows programmatic access to DOI name resolution using HTTP.

Basically, I don't think the DOI foundation thought much about naming the API, so we're left with this kind of long name. You could probably remove "REST" if you wanted.

"DOI Proxy Server API" and "DOI Proxy Server REST API" both pull up the right results on Google for me.

> Will we lose anything if we remove them and add this to the Manubot documentation instead? We would need to move the pandoc-citeproc reference elsewhere.

I do think we lose something, because this is an innovative use of the CSL JSON schema that is a feature that is not widely implemented. I think it is a mistake to not mention major technical innovations behind Manubot to keep the manuscript simple. The manuscript should contain technical details, and I think it is okay for a reader to skip over sentences they don't understand. For example, when I read a biology paper, there are many terms and sentences in the Methods I don't understand. But I think it's probably still better that these sentences are there for the people who do understand them.

content/02.main-text.md
agitter: That is certainly possible, but it isn't the contribution workflow we describe in the paper or recommend.  I'd like to advertise the main reasons to use Manubot here, and I think this is one of them.  Collaborative writing without a review system can get confusing to track as the number of authors grows.

Is there a better and more accurate way to phrase this?  Or should we add the detail in the contribution workflow section that other workflows are possible and this is a project-specific choice?

dhimmel: What about:

> Because manuscripts are created from GitHub repositories, Manubot supports a workflow where all edits are reviewed and discussed, ensuring that the collaborative text has a cohesive style and message and that authors receive precise credit for their work.

dhimmel: > should we add the detail in the contribution workflow section that other workflows are possible and this is a project-specific choice?

I think a short note on this topic is also a good idea. I'd venture that perhaps close to 50% of the manubot manuscripts in existence do not use a PR model and instead primarily commit directly to their master branch. In fact, some don't even ever push the repo to GitHub it seems.

Perhaps we could add a paragraph like:

```
While we developed Manubot with collaborative writing in mind, it can also be helpful for individuals preparing scholarly documents.
In some cases, authors may choose to make their changes directly to the `master` branch, forgoing pull requests and the associated review.
This workflow retains many of the benefits, such as transparent history, automation, and allowing proposed changes from outside contributors.
In cases where outside contributions are unwanted, authors can disable pull requests on GitHub.
See [Existing manuscripts] for examples of a range of contribution workflows and Manubot use cases.
```

@agitter feel free to take any parts of that (or I can open a separate PR)?

agitter: I like both of you quoted text blocks above.  I can work them into this PR tonight.

You're right that we need to clarify that Manubot can be used in many different modes and is even useful for single authors.

agitter:
@dhimmel this is ready for re-review.

> I do think we lose something

You convinced me.  I left it in.

content/02.main-text.md
```diff
@@ -120,14 +125,20 @@ Alternatively, a reviewer can post feedback with a pseudonymous GitHub account o
 Authors can elect to respond to reviews in the GitHub issues or a public [response letter](https://github.com/greenelab/meta-review/blob/813f461fe719d8fe7bcda78871cfad106f83dfe7/content/response-to-reviewers.md), creating open peer review.
 ** TODO: update `response-to-reviewers.md` URL once it is finalized **
 
+Although we developed Manubot with collaborative writing in mind, it can also be helpful for individuals preparing scholarly documents.
+Authors may choose to make their changes directly to the `master` branch, forgoing pull requests and reviews.
+This workflow retains many of Manubot's benefits, such as transparent history, automation, and allowing outside contributors to propose changes.
+In cases where outside contributions are unwanted, authors can disable pull requests on GitHub.
+See the [existing manuscripts](#existing-manuscripts) for examples of the range of contribution workflows and Manubot use cases.
```
dhimmel: `(#existing-manuscripts)` can be removed and this should still work. See [Pandoc's](https://pandoc.org/MANUAL.html#heading-identifiers) `implicit_header_references`

content/02.main-text.md
```diff
@@ -19,6 +19,11 @@ The Deep Review and other studies that subsequently adopted the Manubot platform
 However, inviting wide authorship brought many technical and social challenges such as how to fairly distribute credit, coordinate the scientific content, and collaboratively manage extensive reference lists.
 The manuscript writing process we developed using the Markdown language, the GitHub platform, and our new Manubot tool for automating manuscript generation addresses these challenges.
 
+Manubot supports citations by adding a standard identifier like a Digital Object Identifier (DOI) or PubMed Identifier (PMID) directly in the text so that large groups of authors do not have to coordinate reference lists.
```
dhimmel: "persistent identifier" is probably a more standard term than "standard identifier"

dhimmel (approved):
Nice. Two comments for @agitter to address at his discretion.

agitter:
I made those two additional changes and will merge once the build passes.

Issue #195, opened 2019-04-08
Authorea corrections to writing platform table


agitter:
Alberto from Authorea reviewed our collaborative writing platform table, and we have some corrections to make.  The comments are Hypothesis annotations on this version of the manuscript: https://greenelab.github.io/meta-review/v/dd0882978b274734f4e2197f096678f644cceaca/#tbl:platforms

We can discuss any of these where we need clarification.  We should also review other parts of the text that are relevant, such as our sections on contribution tracking and citation by identifier.

slochower:
We have "Character-level provenance for text" marked at "No" (with brief explanation). Authorea commented and said:

> Yes. There is character-level provenance as every edit is logged in the Git repository behind each article at every save event (every second).

While I have no reason absolutely no reason to cast doubt on this, if I log in to my dummy Authorea manuscript that I used to explore the features last week, and click on the "Version History" tool, I can see a handful of timestamped edits. Moving between the edits sometimes shows me entire words or phrases that I added (some of which is gibberish I was just typing to experiment). Is it possible that Authorea tracks character level provenance but this is not shown to users by default? Or maybe it requires a level of sophistication with the tools that I don't have. I am admittedly a novice Authorea user. But I would say this raises a question of what we are getting across in the table: what is technically done (or technically possible) or what a user sees. Is this distinction clear? (I think this comment would also stand for Overleaf -- an advanced user can probably user the `git` client to synchronize every commit while a web-interface-only user will probably only get certain labeled or time-stamped versions.)

![image](https://user-images.githubusercontent.com/5393995/55735627-90e04100-59d6-11e9-8957-038b140649aa.png)

slochower:
On WYSIWYG: it seems more like Rich Text than true "what you see is what you get" to me, because if you compare the "editing mode" (my term) in the screenshot above and then export a manuscript, the manuscript subjectively *looks* different. True WYSIWYG, in my opinion, should also accurately reflect margins and line breaks, text justification, etc.

[Scientific Paper.pdf](https://github.com/greenelab/meta-review/files/3055028/Scientific.Paper.pdf)

dhimmel:
@slochower did you set up Authorea's GitHub integration? I'd like to see the git repository to learn more about what exactly the commit history contains?

dhimmel:
I'm copying all of Alberto's comments below for reference

### Character-level provenance for text

> **No (versions tracked by day)**
[...]

PR #196, opened 2019-04-08
Collaborative writing platforms section & table

dhimmel:
Closes https://github.com/greenelab/meta-review/issues/195

### DIFF
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -29,10 +29,11 @@ These and other features provide support an open collaborative writing process t
 There are many existing collaborative writing platforms (Table @tbl:platforms) [@doi:10.1038/514127a].
 In general, platforms with "what you see is what you get" (WYSIWYG) editors, such as Microsoft Word or Google Docs, require the least technical expertise to use.
 On the flip side, WYSIWIG platforms can be difficult to customize and incorporate into automated computational workflows.
-Traditionally, LaTeX has been used for these needs, since documents are written in plain text and the system is open source and customizable.
+Traditionally, LaTeX has been used for these needs, since documents are written in plain text and the system is open source and extensible.
 Rendering LaTeX documents requires specialized software, but webapps like [Overleaf](https://www.overleaf.com/) now enable collaborative authoring of LaTeX documents.
 Nonetheless, LaTeX-based systems are limited in that PDF (or similar) is the only fully supported output format.
-Alternatively, [Authorea](https://www.authorea.com/) is a collaborative writing webapp whose primary output format is HTML, but allows authors to write sections in Markdown, a limited subset of LaTeX, or their WYSIWIG HTML editor.
+Alternatively, [Authorea](https://www.authorea.com/) is a collaborative writing webapp whose primary output format is HTML.
+Authorea allows authors to write sections in Markdown, a limited subset of LaTeX, or their WYSIWIG HTML editor.
 
 | Feature | Manubot | Authorea | Overleaf v2 | Google Docs + Paperpile | Word Online<sup>1</sup> | Markdown on GitHub |
 |---|---|---|---|---|---|---|
@@ -65,8 +66,7 @@ The GitHub Sync feature attributes all edits to the project owner.
 
 Existing platforms work well for editing text and are widely used for scholarly writing.
 However, they often lack features that are important for open collaborative writing, such as versatile version control and multiple permission levels.
-Table @tbl:platforms compares several existing platforms.
-None of these platforms besides Manubot offer the ability to address thematically related changes together and enable multiple authors to iteratively refine proposed changes.
+For example, Manubot is the only platform listed in Table @tbl:platforms that offers the ability to address thematically related changes together and enables multiple authors to iteratively refine proposed changes.
 
 ## Manubot contribution workflow {#contribution-workflow}
 
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -26,17 +26,13 @@ These and other features provide support an open collaborative writing process t
 
 ## Collaborative writing platforms
 
-There are many existing collaborative writing platforms [@doi:10.1038/514127a].
+There are many existing collaborative writing platforms (Table @tbl:platforms) [@doi:10.1038/514127a].
 In general, platforms with "what you see is what you get" (WYSIWYG) editors, such as Microsoft Word or Google Docs, require the least technical expertise to use.
 On the flip side, WYSIWIG platforms can be difficult to customize and incorporate into automated computational workflows.
 Traditionally, LaTeX has been used for these needs, since documents are written in plain text and the system is open source and customizable.
-Rendering LaTeX documents requires specialized software, but webapps like [Overleaf](https://www.overleaf.com/) now enable collaborative authoring of LaTeX documents via a webapp.
-Nonetheless, LaTeX-based systems are limited in that PDF is the only fully supported output format.
+Rendering LaTeX documents requires specialized software, but webapps like [Overleaf](https://www.overleaf.com/) now enable collaborative authoring of LaTeX documents.
+Nonetheless, LaTeX-based systems are limited in that PDF (or similar) is the only fully supported output format.
 Alternatively, [Authorea](https://www.authorea.com/) is a collaborative writing webapp whose primary output format is HTML, but allows authors to write sections in Markdown, a limited subset of LaTeX, or their WYSIWIG HTML editor.
-Collaborative writing tools ideally offer version control, multiple permission levels, or other functionality to support multi-author document editing.
-Existing platforms work well for editing text and are widely used for scholarly writing.
-However, they lack features that are important for open writing, such as managing a collaborative manuscript and attributing precise credit among many authors (Table @tbl:platforms).
-None of these platforms offer the ability to address thematically related changes together and enable multiple authors to iteratively refine proposed changes.
 
 | Feature | Manubot | Authorea | Overleaf v2 | Google Docs + Paperpile | Word Online<sup>1</sup> | Markdown on GitHub |
 |---|---|---|---|---|---|---|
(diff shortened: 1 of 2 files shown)


content/02.main-text.md
```diff
@@ -29,20 +29,20 @@ Existing platforms work well for editing text and are widely used for scholarly
 However, they lack features that are important for open writing, such as managing a collaborative manuscript and attributing precise credit among many authors (Table @tbl:platforms).
 None of these platforms offer the ability to address thematically related changes together and enable multiple authors to iteratively refine proposed changes.
 
-| Feature | Manubot | Authorea + BibTeX | Overleaf v2 + BibTeX | Google Docs + Paperpile | Word Online<sup>1</sup> | Markdown on GitHub |
```
dhimmel, line 37: I removed "+ BibTeX" because I think that in the case of authorea it may be inaccurate and in the case of Overleaf it may be assumed. 

It seems like BibTex refers to a program and less so the `.bib` file format. https://tex.stackexchange.com/a/299286 And I am not sure if Authorea uses the BibTeX program to render their citations, but I think it is likely that they do not.

content/02.main-text.md
```diff
@@ -29,20 +29,20 @@ Existing platforms work well for editing text and are widely used for scholarly
...
 | Continuous integration testing | Yes | No | No | No | No | No |
 | Multi-participant conversation for changes | Yes | No<sup>2</sup>  | No<sup>2</sup> | No<sup>2</sup> | No<sup>2</sup> | Yes |
-| Character-level provenance for text | Yes | No (versions tracked by day) | No<sup>3</sup> | Requires manual inspection of history | Not after changes are accepted | Yes |
+| Character-level provenance for text | Yes | Yes | No<sup>3</sup> | Requires manual inspection of history | Not after changes are accepted | Yes |
 | Bibliography management | Yes | Yes | Yes | Yes | No, requires the Word desktop application | No |
-| Citation by identifier | Yes | No | No | No | No | No |
+| Citation by identifier | Yes | Yes | No | No | No | No |
 | Editing software | Any text editor | Web interface | Web interface | Web interface | Web interface | Any text editor |
-| Document format | Markdown | LaTeX | LaTeX | Proprietary | Proprietary | Markdown |
+| Document format | Markdown | HTML | LaTeX | Proprietary | Proprietary | Markdown |
```
dhimmel, line 48: It seems like the primary output format is HTML although the input is a mix of HTML exposed via a WYSIWIG-editor, markdown cells and latex cells.

content/02.main-text.md
slochower: Agree

content/02.main-text.md
slochower: Agree

agitter (commented):
These changes to the table all look good to me.  There are a few other parts of the manuscript we can change as well to keep things consistent:

> Others are LaTeX-based systems for technical writing [6], such as Overleaf and Authorea.

If Authorea is not LaTeX-based, we need to rephrase this.

The Table 1 caption says
> We assessed features on June 15, 2018 using the free version of each platform and updated our assessment on April 3, 2019 to add Overleaf v2 and the features in the bottom three rows.

Should we update the date of the Authorea evaluation as well?  I'm not sure how many of these features may have changed since June 2018.

The `Citation by identifier` section says
>Manubot includes an additional layer of citation processing, currently unique to the system.

I believe that phrasing is still accurate even if Authorea has its own citation by identifier approach, right?

dhimmel:
I moved the "Collaborative writing platforms" content to its own section rather than in "Manubot contribution workflow". I also rewrote the start of that paragraph to provide a little more context and detail on existing platforms.

content/02.main-text.md
```diff
@@ -24,36 +24,40 @@ When text is changed, Manubot automatically updates the manuscript's web page so
...
-Some provide "what you see is what you get" (WYSIWYG) interfaces and support Microsoft Word or similar document formats.
-Others are LaTeX-based systems for technical writing [@doi:10.1038/514127a], such as [Overleaf](https://www.overleaf.com/) and [Authorea](https://www.authorea.com/).
-These tools ideally offer version control, multiple permission levels, or other functionality to support multi-author document editing.
+## Collaborative writing platforms
+
+There are many existing collaborative writing platforms [@doi:10.1038/514127a].
+In general, platforms with "what you see is what you get" (WYSIWYG) editors, such as Microsoft Word or Google Docs, require the least technical expertise to use.
+On the flip side, WYSIWIG platforms can be difficult to customize and incorporate into automated computational workflows.
+Traditionally, LaTeX has been used for these needs, since documents are written in plain text and the system is open source and customizable.
+Rendering LaTeX documents requires specialized software, but webapps like [Overleaf](https://www.overleaf.com/) now enable collaborative authoring of LaTeX documents via a webapp.
```
slochower: Probably don’t need the second webapp.

content/02.main-text.md
```diff
@@ -24,36 +24,40 @@ When text is changed, Manubot automatically updates the manuscript's web page so
...
+Nonetheless, LaTeX-based systems are limited in that PDF is the only fully supported output format.
```
slochower: I might say “usually” here. For a long time DVI was used as the primary output.

content/02.main-text.md
```diff
@@ -24,36 +24,40 @@ When text is changed, Manubot automatically updates the manuscript's web page so
...
+Alternatively, [Authorea](https://www.authorea.com/) is a collaborative writing webapp whose primary output format is HTML, but allows authors to write sections in Markdown, a limited subset of LaTeX, or their WYSIWIG HTML editor.
+Collaborative writing tools ideally offer version control, multiple permission levels, or other functionality to support multi-author document editing.
```
slochower: I think you can merge this sentence with the following sentence to ease the transition because they are both declarative about other tools, in general.

content/02.main-text.md
dhimmel: What about adding the following qualification:

> Nonetheless, LaTeX-based systems are limited in that PDF (including PDF predecessors) is the only fully supported output format.

dhimmel: Or perhaps `(or similar)`

slochower: I like similiar. It also covers things like .dvju and to me, it implies static and non-interactive which is where HTML output shines. I think LaTeX to HTML usually goes via PDF, so there is little to no interactivity.

content/02.main-text.md
```diff
@@ -24,36 +24,36 @@ When text is changed, Manubot automatically updates the manuscript's web page so
...
-These tools ideally offer version control, multiple permission levels, or other functionality to support multi-author document editing.
-Existing platforms work well for editing text and are widely used for scholarly writing.
-However, they lack features that are important for open writing, such as managing a collaborative manuscript and attributing precise credit among many authors (Table @tbl:platforms).
-None of these platforms offer the ability to address thematically related changes together and enable multiple authors to iteratively refine proposed changes.
+There are many existing collaborative writing platforms (Table @tbl:platforms) [@doi:10.1038/514127a].
+In general, platforms with "what you see is what you get" (WYSIWYG) editors, such as Microsoft Word or Google Docs, require the least technical expertise to use.
+On the flip side, WYSIWIG platforms can be difficult to customize and incorporate into automated computational workflows.
+Traditionally, LaTeX has been used for these needs, since documents are written in plain text and the system is open source and customizable.
```
agitter: ` and customizable` feels redundant to me because `these needs` refers to customize and automate in the prior sentence.

content/02.main-text.md
```diff
@@ -24,36 +24,36 @@ When text is changed, Manubot automatically updates the manuscript's web page so
...
+Rendering LaTeX documents requires specialized software, but webapps like [Overleaf](https://www.overleaf.com/) now enable collaborative authoring of LaTeX documents.
+Nonetheless, LaTeX-based systems are limited in that PDF (or similar) is the only fully supported output format.
+Alternatively, [Authorea](https://www.authorea.com/) is a collaborative writing webapp whose primary output format is HTML, but allows authors to write sections in Markdown, a limited subset of LaTeX, or their WYSIWIG HTML editor.
```
agitter: Maybe split into two sentences:
> primary output format is HTML.  It allows authors

Are we sure it only supports a limited subset of LaTeX?  I haven't checked myself, so I'll assume you tested this.

content/02.main-text.md
```diff
@@ -63,6 +63,13 @@ Edits made by distinct authors may be attributed to a single author.
 The GitHub Sync feature attributes all edits to the project owner.
 {#tbl:platforms}
 
+Existing platforms work well for editing text and are widely used for scholarly writing.
+However, they often lack features that are important for open collaborative writing, such as versatile version control and multiple permission levels.
+Table @tbl:platforms compares several existing platforms.
```
agitter: This sentence seems redundant to me.  Can we cut it and add the table reference elsewhere in this paragraph in a parenthetical if you think we need it?

content/response-to-reviewers.md
```diff
@@ -173,7 +173,7 @@ Same for inline comments and diff colorization.
 
 We added `WYSIWYG mode`, `Inline comments`, and `Viewing changes` as new features to the table that compares collaborative writing platforms.
 In addition, we updated the table to use Overleaf v2 instead of Overleaf v1, which was discontinued.
-We discussed these changes in [GH147](https://github.com/greenelab/meta-review/issues/147) and edited the table in [GH188](https://github.com/greenelab/meta-review/pull/188).
+We discussed these changes in [GH147](https://github.com/greenelab/meta-review/issues/147) and edited the table in [GH188](https://github.com/greenelab/meta-review/pull/188) and [GH196](https://github.com/greenelab/meta-review/pull/196).
```
agitter, line 184: Can you also please list this PR in the preceding reviewer comment about `they lack sufficient features for managing a collaborative editing`?  Your new paragraph better explains the difference between Manubot and other tools and gives existing tools their due credit.

agitter (changes requested):
These are helpful changes.  I have a few more suggestions.

content/02.main-text.md
dhimmel: From https://support.authorea.com/en-us/article/how-do-i-include-a-latex-package-1bjtigr/:

> Authorea features [LaTeXML](http://dlmf.nist.gov/LaTeXML/), the most powerful way to render LaTeX notation on the web. [This](https://github.com/Authorea/LaTeXML/tree/master/lib/LaTeXML/Package) is an up to date list of all latex packages you can use on Authorea.

agitter (approved):

Issue #145, opened 2019-02-08
Reviewer 2 comment 16


agitter:
> `we made the GitHub repository public under a Creative Commons Attribution License.` → The link points to CC-BY 4.0 International, it would be good to have this information in the name of the link (or we have to click to check which kind CC you used).

PR #199, opened 2019-04-09
Reduce span of hyperlinked text

dhimmel:
[Prior](https://greenelab.github.io/meta-review/v/675980c23332d6a62b8849c455ebc7915d6683e0/#introduction) to this PR, the amount of text linked for the license was a bit excessive IMO:

> image

I believe this should still address the reviewer comment in https://github.com/greenelab/meta-review/issues/145.


dhimmel:
I removed "4.0 International" from the long name because they are a bit redundant IMO with the parenthesized statement. I think most individuals who care about which version of CC BY license was used will know that 4.0 corresponds to International. Also, this sentence is not the actual licensing statement of the deep review, so precision is not actually required. Hyperlink title contains most verbose description for those viewing the HTML.

agitter (approved):
This is a nice compromise.

PR #205, opened 2019-04-10
Use Athena instead of WeasyPrint for PDF creation if possible

### LINKED ISSUE #202: Figure 2 cut off in PDF
The right side of Figure 2, the deep review contributions plot, is truncated in the current PDF.

@vincerubinetti @dhimmel any ideas as to what could be causing this?

dhimmel:
Closes https://github.com/greenelab/meta-review/issues/202
Refs https://github.com/manubot/rootstock/issues/170

```
Co-authored-by: Vincent Rubinetti <vince.rubinetti@gmail.com>
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 dist: trusty
 sudo: false
 language: generic
+services:
+  - docker
 branches:
   only:
     - master
--- a/build/build.sh
+++ b/build/build.sh
@@ -81,7 +81,7 @@ fi
 if [ "$BUILD_PDF" != "false" ] && [ -n "$DOCKER_EXISTS" ]; then
   echo "Exporting PDF manuscript using Docker + Athena"
   if [ -d output/images ]; then rm -rf output/images; fi  # if images is a directory, remove it
-  cp --recursive --dereference content/images output/
+  cp -r -L content/images output/
   docker run \
     --rm \
     --volume `pwd`/output:/converted/ \
--- a/build/build.sh
+++ b/build/build.sh
@@ -51,9 +51,12 @@ pandoc --verbose \
   --output=output/manuscript.html \
   $INPUT_PATH
 
+# Return null if docker command is missing, otherwise return path to docker
+DOCKER_EXISTS=`command -v docker`
+
 # Create PDF output (unless BUILD_PDF environment variable equals "false")
-if [ "$BUILD_PDF" != "false" ]; then
-  echo "Exporting PDF manuscript"
+if [ "$BUILD_PDF" != "false" ] && [ -z "$DOCKER_EXISTS" ]; then
+  echo "Exporting PDF manuscript using WeasyPrint"
   if [ -L images ]; then rm images; fi  # if images is a symlink, remove it
   ln -s content/images
   pandoc \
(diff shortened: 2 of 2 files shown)


dhimmel:
For reference, here are the [`athenapdf`](https://github.com/arachnys/athenapdf) options:

```
>>> athenapdf --help
Usage: athenapdf [options] <URI> [output]

convert HTML to PDF via stdin or a local / remote URI

Options:
  -V, --version                  output the version number
  --debug                        show GUI
  -T, --timeout <seconds>        seconds before timing out (default: 120)
  -D, --delay <milliseconds>     milliseconds delay before saving (default: 200)
  -P, --pagesize <size>          page size of the generated PDF (default: A4) (default: "A4")
  -M, --margins <marginsType>    margins to use when generating the PDF (default: standard) (default: "standard")
  -Z --zoom <factor>             zoom factor for higher scale rendering (default: 1 - represents 100%)
  -S, --stdout                   write conversion to stdout
  -A, --aggressive               aggressive mode / runs dom-distiller
  -B, --bypass                   bypasses paywalls on digital publications (experimental feature)
  -H, --http-header <key:value>  add custom headers to request (default: [])
  --proxy <url>                  use proxy to load remote HTML
  --no-portrait                  render in landscape
  --no-background                omit CSS backgrounds
  --no-cache                     disables caching
  --ignore-certificate-errors    ignores certificate errors
  --ignore-gpu-blacklist         Enables GPU in Docker environment
  --wait-for-status              Wait until window.status === WINDOW_STATUS (default: wait for page to load)
  -h, --help                     output usage information
```

dhimmel:
From [build](https://travis-ci.org/greenelab/meta-review/builds/518373156#L827) of https://github.com/greenelab/meta-review/pull/205/commits/c03d9f03afa1d9f122bb76eb28e792f034551ebb:

```
Exporting PDF manuscript using Docker + Athena
Unable to find image 'arachnysdocker/athenapdf:2.16.0' locally
2.16.0: Pulling from arachnysdocker/athenapdf
Status: Downloaded newer image for arachnysdocker/athenapdf:2.16.0
ATTENTION: default value of option force_s3tc_enable overridden by environment.
Converted 'file:///converted/manuscript.html' to PDF: 'manuscript.pdf'
PDF Conversion: 7860.561ms
```

dhimmel:
Here's the ouput PDF built locally: [manuscript.pdf](https://github.com/greenelab/meta-review/files/3064787/manuscript.pdf).

Fixes #202. CC @vincerubinetti

agitter (approved):
It build changes and output PDF look great to me.  We typically only make changes like this upstream in Rootstock.  Do you want to do more testing before adding this support there?

dhimmel:
> We typically only make changes like this upstream in Rootstock. Do you want to do more testing before adding this support there?

I am thinking meta-review will be a bit of testing and then we can decide how we want to port it to rootstock.

agitter:
That works for me.  Merge when ready.

slochower:
Just piping in to say I tried Athena on one of my other manuscripts and it looks **much better**. I support this 👍!

(Edit: also WeasyPrint was timing out and when I manually updated to the latest release, I got a segmentation fault... and on Travis some of the equations took too long to render so it didn't look so great: so there were a few problems that Athena has solved.)

slochower:
Note that the page size is not 8.5" x 11".

image

I was looking here to see whether metadata is set for the PDF, which it is not. But I do not think we will be able to do that directly with Athena PDF.

https://github.com/arachnys/athenapdf/issues/158

It would be nice if we could at least set the title to something other than "manuscript.pdf". The manuscript title would make a much more informative browser tab.

slochower:
For what it's worth, it is easy to update the metadata using `exiftool`:
```
exiftool -Title="This is the Title" -Author="This is the author" manuscript\ \(2\).pdf
```
Which could be installed on Travis using `apt`. I'm not sure we'd want to specify author for multi-author manuscripts, but title would be nice in the metadata and/or by renaming the file. (I realize renaming presents problems with linking to the manuscript.pdf URL, especially if the title changes during development.)

PR #213, opened 2019-04-11
Remove v1 suffix

slochower:


dhimmel:
@slochower I actually think we should specify v2 in the citation. This allows us to update the figshare in the future without making the manuscript use a different video embed than the citation refers to.

slochower:
Okay, fair point. I guess we could totally change what the video says...

dhimmel:
> Okay, fair point. I guess we could totally change what the video says...

Not only that, the embed links need to be manually updated... they refer to a specific file from a specific version. So we might as well have all the links be versioned rather than some versioned, some unversioned.

dhimmel (approved):

PR #211, opened 2019-04-11
Himmelstein's proofread

dhimmel:
- [x] update keywords
- [x] CSL updates in https://github.com/greenelab/meta-review/pull/203#issuecomment-481797833

PR #200, opened 2019-04-09
Track changes using pandiff

dhimmel:
Closes https://github.com/greenelab/meta-review/issues/198

Work in progress.

PR #216, opened 2019-04-12
Continuous publishing/publication usage

dhimmel:
Refs https://github.com/greenelab/meta-review/pull/211#discussion_r274972195

### DIFF
--- a/build/assets/style.csl
+++ b/build/assets/style.csl
@@ -16,6 +16,10 @@
   <macro name="author">
     <names variable="author">
       <name initialize="false" initialize-with="" et-al-min="12" et-al-use-first="10" et-al-use-last="true"/>
+      <substitute>
+        <names variable="editor"/>
+        <text macro="venue"/>
+      </substitute>
     </names>
   </macro>
   <macro name="venue">
--- a/content/02.main-text.md
+++ b/content/02.main-text.md
@@ -417,7 +417,7 @@ Manubot provides an ideal platform for perpetual reviews [@arxiv:1502.01329; @ta
 
 Concepts for the future of scholarly publishing extend beyond collaborative writing [@doi:10.22541/au.149693987.70506124; @tag:techblog-brown].
 Bookdown [@doi:10.1201/9781315204963] and Pandoc Scholar [@doi:10.7717/peerj-cs.112] both extend traditional Markdown to better support publishing.
-Examples of continuous integration to automate manuscript generation include [gh-publisher](https://github.com/ewanmellor/gh-publisher) and Continuous Publishing [@url:http://blog.martinfenner.org/2014/03/10/continuous-publishing/], which was used to produce the book Opening Science [@doi:10.1007/978-3-319-00026-8].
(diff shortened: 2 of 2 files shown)


dhimmel:
I'd like to choose whether we should use the "continuous publishing" or "continuous publication" in this PR.

Copied from https://github.com/greenelab/meta-review/pull/211#discussion_r274972195:

continuous publishing has [674 Google Scholar results](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C39&q=%22continuous+publishing%22&btnG=) versus [4,400](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C39&q=%22continuous+publication%22&btnG=) for "continuous publication".

However, much of the usage of "continuous publication" is regarding journals moving away from issue-based publishing. [Example](https://doi.org/10.1259/bjr.20130252)

> Continuous publication focuses on publishing individual articles as opposed to the more traditional issue-based publishing. Once an article is in its final form, after editing and proofreading, it is immediately published online as the version of record straight into an issue. There is no delay caused by waiting for a group of manuscripts to be gathered together for a given issue. This drip feeding of articles means that issues are built up gradually throughout the month.

Continuous publishing is used by Fenner's [post](http://blog.martinfenner.org/2014/03/10/continuous-publishing/) to describe Manubot's use case.

Hence, I feel like we're slightly coining the term, so we should probably just go with what we think is most accurate?

***

@mfenner do you remember why you selected the "continuous publishing" versus "continuous publication" and do you have a current recommendation? We are referring to the process that a manuscript gets automatically rebuild and deployed using CI whenever its git-tracked source code changes.

dhimmel:
In favor of continuous publication, is:

- continuous delivery not continuous delivering
- continuous integration not continuous integrating
- continuous deployment not continuous deploying

[Another link](http://doi.org/10.1629/2048-7754.61 "350 years at the cutting edge of scientific publishing – the Royal Society moves to continuous publication") regarding the existing usage of this term for the less radical unbundled publication of journal articles.

build/assets/style.csl
```diff
@@ -16,6 +16,10 @@
   <macro name="author">
     <names variable="author">
       <name initialize="false" initialize-with="" et-al-min="12" et-al-use-first="10" et-al-use-last="true"/>
+      <substitute>
+        <names variable="editor"/>
+        <text macro="venue"/>
+      </substitute>
```
dhimmel, line 22: From http://docs.citationstyles.org/en/stable/specification.html#substitute

> Substituted variables are suppressed in the rest of the output to prevent duplication. An example, where an empty “author” name variable is substituted by the “editor” name variable, or, when no editors exist, by the “title” macro:

mfenner:
I don't think I thought much about continuous publishing vs. continuous publication when I wrote that in 2014.

dhimmel:
@agitter ready for review. After this, I can merge the changes to produce diffs in https://github.com/greenelab/meta-review/pull/200

agitter (approved):
Looks good.  One minor note, you meant the diffs in #200 right?  Once this is merged and builds, I'll take ~5 minutes to scan everything and check for last minute problems.

dhimmel:
The effect of the CSL change is shown at https://github.com/greenelab/meta-review/commit/8be2fc76ebf2d7bbe45ed3e4bd145b8aba62c8e5.

PR #240, opened 2019-06-10
increase resolution of plugin screenshots

vincerubinetti:
plugins

To 600dpi. 4500px by 3800px.


dhimmel:
Nice! Looks a lot better, especially when zooming in.

Rather than put all the subfigures in the `content/figures` directory, can we put them in `analyses/plugin-screenshots`. We could also move `plugins.psd` there. I think it is possible that having too much stuff in `content/images` could cause problems because that directory gets duplicated so many times. The repository shouldn't get too big, but instead cloning the repo onto a filesystem may become more challenging.

We can look into the conflicts in the office tomorrow.

agitter:
The new higher-resolution figure looks fantastic.  I'm ready to merge this after you decide where to move the subfigures and resolve the conflicts.

dhimmel (approved):
Looks good to me.

Issue #233, opened 2019-05-31
Reference format


agitter:
> Your references need to be formatted according to the PLOS style. PLOS uses the numbered citation (citation-sequence) method and first six authors, et al. Authors should be listed as "Last name, First Initial". If using EndNote, please use the following style file: https://endnote.com/style_download/plos-public-library-of-science-all-journals/

See https://editor.citationstyles.org/styleInfo/?styleId=http%3A%2F%2Fwww.zotero.org%2Fstyles%2Fplos-computational-biology

dhimmel:
See https://github.com/greenelab/meta-review/pull/228

dhimmel:
More info at https://journals.plos.org/ploscompbiol/s/submission-guidelines#loc-references

agitter:
One formatting question I'm still unclear about is whether they will take all of our inline links throughout the text.  Do we need to reformat those as URL citations?

dhimmel:
> Do we need to reformat those as URL citations?

I'm opposed to polluting our reference section with hyperlinks. Is there any indication that PLOS CB want us reformat hyperlinks as citations?

I would say we just wait to the proof stage to see if they make it though. I think their system is technically capable of publishing hyperlinks, although there is a chance they get removed by the default workflow.

agitter:
No, I haven't seen any indication that they want us to change the hyperlinks.  I checked a few papers from their current home page and saw that they write out the URL for hyperlinks:

https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1007044
- `OpenCV ver. 3.3.1 (https://opencv.org/).`
- `ImageJ (https://imagej.nih.gov)`

https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1007056
- `Los Alamos HIV Database (http://www.hiv.lanl.gov/)`
- `PyMol (https://pymol.org/2/)`

I'm okay leaving our links in their current format for the resubmission.

PR #241, opened 2019-06-11
Cleanup PLOS styled references

dhimmel:
Style references according to https://github.com/greenelab/meta-review/issues/233.


dhimmel:
I opened https://github.com/citation-style-language/styles/issues/4138 to discuss how `[Internet]` is applied to many references like preprints and books, which seems to contradict the [PLOS guidelines](https://journals.plos.org/ploscompbiol/s/submission-guidelines#loc-references). Also opened https://github.com/citation-style-language/styles/issues/4140 regarding publishers being shown for journal articles.

dhimmel:
See the references in this PDF as of https://github.com/greenelab/meta-review/pull/241/commits/d2eed428e31f104fbbe35c5bb0029b599b10c96d: [manuscript.pdf](https://github.com/greenelab/meta-review/files/3278274/manuscript.pdf).

Comment with any issues or imperfections that are not mentioned above.

agitter:
One side effect of the PLOS reference style is that some of our links to the first use of the reference break:
image

There is no longer a `31` in the text above to link back to:
image

Thats unavoidable but a good reason to switch back to our default style after resubmission.

agitter:
Here are some other issues or potential issues:
- Our arXiv references don't match their example.  We don't include Preprint or the Cited date.  We do include the Report No. and other text.
- They do not want personal communication in the reference list (reference 21).
- 48 has been published in a journal since our original citation.
- The math in the title of 54 doesn't translate to plain text well.
- References 7, 8, 50, 52, and 58 have editors listed for published articles.
- Our syntax for figshare doesn't match their example `Database: figshare`

dhimmel:
> They do not want personal communication in the reference list (reference 21).

Good observation that the [guidelines](https://journals.plos.org/ploscompbiol/s/submission-guidelines#loc-references) state:

> Personal communications (these should be supported by a letter from the relevant authors but not included in the reference list)

However, in [our case](https://greenelab.github.io/meta-review/v/040638d25b9a45ad3d060afc9689a17f4c3a3045/#ref-l625msOK) the citation / reference is just an example. It would distort the interpretation of the table to not include the citation in the references, right? Therefore, perhaps we can preemptively note that this is a special case.

I will fix the title of [ref 54](https://greenelab.github.io/meta-review/v/040638d25b9a45ad3d060afc9689a17f4c3a3045/#ref-15YfJWkEd). However the rest of the items seem to require CSL changes that could be time consuming. Therefore, we should consider the following options:

+ leaving some of the issues unremidied. It's not clear to me that the journal has even put much consideration into the formatting of their references. They may not enforce the example format or they may have their staff correct it.
+ manually making these changes to the DOCX. If we keep a version with tracked changes, we would also be able to see what we did (in case we have to do it again).

agitter:
For the `raw` citation example, should we switch to a different type of example besides personal communication?  Especially now that Dongbo is an author, we don't necessarily need to use that one.

I suggest that we change the Kipoi reference (48) to https://doi.org/10.1038/s41587-019-0140-0.  There is only one version of the preprint, and the final published version of the manuscript has a new title and some notable content changes.

> leaving some of the issues unremedied

This is my preference for everything else.  I expect that most authors don't manually fix reference formatting if the mistakes come from their reference manager or citation style file.

dhimmel:
> For the raw citation example, should we switch to a different type of example besides personal communication? Especially now that Dongbo is an author, we don't necessarily need to use that one.

Suggestions? The difficulty was that we wanted a type of citeable work that doesn't have any existing persistent identifiers. So this has to be something private / not on the internet. Now we could use a `raw:` citation for anything, but then we may deviate from citaiton-by-identifier best practice.

agitter:
I revisited the PLOS reference guidelines to see if there is anything we could substitute for the personal communication `raw` citation example.  Everything they place in the references should have a standard identifier available except for print media without an online counterpart and `Multimedia (videos, movies, or TV shows)`.  That may be a stretch to cite in this manuscript unless someone has a clever idea.

agitter:
Has anyone seen [_Paywall_](https://paywallthemovie.com/)?  That could work as a movie citation.

dhimmel:
> Has anyone seen Paywall? That could work as a movie citation.

Yes, it is OKAY but not GREAT. It's intentions are good, but the piecing together of interviews leaves the narrative a bit scattered.

Switched in https://github.com/greenelab/meta-review/pull/241/commits/383a6388c0ff5ade9fa4263140a206692f3e9af7.

Reference now renders as:

> 21. Paywall: The Business of Scholarship [Internet]. 2018. Available:
https://paywallthemovie.com/paywall

So may need manual touch up of rendered reference.

agitter (approved):

dhimmel:
@agitter take another look with the new commits (especially example manuscripts section). PDF as of https://github.com/greenelab/meta-review/pull/241/commits/ab27ba102dd2f44c6932f430779d9ed1f6614629: [manuscript.pdf](https://github.com/greenelab/meta-review/files/3286970/manuscript.pdf)

content/02.main-text.md
```diff
@@ -302,14 +302,28 @@ As such, continuous publication via Manubot helped the authors address the error
...
-Manubot is also being used for documents beyond traditional journal publications, such as [research tips](https://benjamin-lee.github.io/deep-rules/), [quality standards](https://indigo-dc.github.io/sqa-baseline/) [@url:https://hdl.handle.net/10261/160086], [grant proposals](https://greenelab.github.io/manufund-2018/), [progress reports](https://greenelab.github.io/czi-hca-report/), [undergraduate research reports](https://zietzm.github.io/Vagelos2017/) [@doi:10.6084/m9.figshare.5346577], [literature reviews](https://slochower.github.io/synthetic-motor-literature/), and lab notebooks.
+Additional research studies are being authored using Manubot, spanning the fields of
+[regulatory](https://vsmalladi.github.io/tfsee-manuscript/)
+[genomics](https://simonvh.github.io/gimmemotifs-manuscript/) [@doi:10.1101/474403],
+[synthetic biology](https://zach-hensel.github.io/low-noise-manuscript/) [@doi:10.1101/515940],
```
agitter: Now published http://doi.org/10.1128/mSphere.00340-19

agitter (commented):
I found one more update for a Manubot-written manuscript.  Everything else looks good.

agitter:
Looks good to me.

Issue #226, opened 2019-05-31
PLOS formatting requests


agitter:
The PLOS Computational Biology team emailed the following message about formatting.  The attachment below contains the specific requests.  I'll create individual issues for each request and use this as the meta issue, as during revision.
- [x] Metadata verification #229
- [x] Provide a .doc or .docx version of your manuscript file.
- [x] Numeric affiliations #230
- [x] Author summary paragraphs #231
- [x] Remove figures from text #232
- [x] Reference format #233
- [x] Figure formatting #234
- [x] Striking image #235

Should I contact them to ask about all of the links in our text?  If they remove them during proofing without telling us, the text will suffer.

> This is the FINAL stage for you to copyedit your manuscript and figure files. We strongly recommend that all authors double check ALL files prior to the final versions being submitted. The version submitted at this stage will be posted online ahead of the final version, unless you have already indicated that you do not wish to post an early version. Changes to the manuscript, statements, author list, references, equations, or figures (including resizing) at the proof stage WILL NOT BE ACCEPTED 

> Thank you for your patience while we evaluated your manuscript, 'Open collaborative writing with Manubot', to ensure it meets our formatting requirements prior to publication. 

> We require you to make some technical changes to your manuscript, which are listed in the attached document. Your manuscript will not proceed to production until you have satisfactorily completed these requests.

> To access your manuscript and complete these changes, please follow this link `<redacted>`. You will find the submission in "Current Task Assignments", and will need to click "Submit Task" in order to upload your corrected file(s).

> Please note, once these technical changes have been completed the corresponding author will receive a typeset proof for review, to ensure errors have not been introduced during production. Additional changes WILL NOT be accommodated at the proof stage and a correction notice will need to be issued after publication if a further change is required and appropriate. 

> It is strongly recommended that ALL authors now check the manuscript and figures, as well as author-provided statements including:
> *	The author list, contributions, and affiliations
> *	Data availability 
> *	Funding
> *	Conflicts of interest

> These statements are also provided for review in the attached document with the required technical changes to your document. This is the FINAL opportunity to make changes to your manuscript, so please do check ALL files thoroughly.

> If you have any questions about these requests, please do get in contact. 

Instructions in the submission system:

[...]

PR #254, opened 2019-08-30
Revert some PLOS formatting requirements

dhimmel:
For more info, see https://github.com/greenelab/meta-review/issues/226#issuecomment-497860986.


agitter (approved):
I like the addition of the link to the changes since publication.  b90923a99208c9152cec7929399ab1b7929cb7e2 seems like an appropriate base instead of the `v4.0` tag.

PR #264, opened 2020-01-27
Hyperlink DOIs to preferred resolver

katrinleinweber:
Same as https://github.com/greenelab/scihub/pull/62 ;-)


agitter:
Thanks for the contribution @katrinleinweber.  We've been using `https://doi.org` instead of `http://dx.doi.org` whenever possible.  These two specific files are archival files; they are not part of the source used to build the Manubot manuscript.  For that reason, we may prefer to leave them in their original form.

@dhimmel what do you think?

dhimmel:
Thanks for looking out for our hyperlinks (:

> These two specific files are archival files; they are not part of the source used to build the Manubot manuscript. For that reason, we may prefer to leave them in their original form.

Yes, we're archiving and creating diffs on files created by PLOS Computational Biology. Changing these links would be a corruption.

